PR #3, opened 2015-04-24
Fixing ordering from Article II item 4

rotilho:
Fixing ordering from Article II item 4


slush0:
This github repository is currently the mirror of official source located at liberland.org. This should be reported to liberland.org instead of fixing here. Anyway, thank you for report!

PR #4, opened 2015-04-24
Remove authority based on 'law of nature'. This could mean anything and ...

joshmh:
...consent of governed is sufficient.

### DIFF
--- a/Liberland-constitution.md
+++ b/Liberland-constitution.md
@@ -4,7 +4,7 @@ Last changed: Thu Apr 23 2015 12:55:22 GMT+0200 (CEST)
 
 > Notice: this is NOT a final version of Constitution, it is just a draft!
 
-*We, the Citizens of the Free Republic of Liberland, in order to establish justice, insure domestic tranquility, provide for the common defense, and secure the blessings of liberty to ourselves and future generations, do ordain and establish the Constitution of the Free Republic of Liberland, hereinafter “the Constitution”, as its supreme law, deriving its just authority from the law of nature and the consent of the governed. Hereby, the Free Republic of Liberland shall be governed by the Public Administration exercising the legislative, executive and judicial power restricted by the Bill of Rights.*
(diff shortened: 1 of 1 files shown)


naconner:
While I agree, I think it can be clear within the Lockean tradition what's meant by 'law of nature.'

However, the problem with this phrase is that it undermines a rational, objective basis for rights, ascribing them ultimately to a source which must be accepted on faith.

> But the idea that rights come from a law of nature created by God is beset with all the same problems as the idea that rights come from God—because it is the same idea, albeit with God’s involvement one step removed.

> If natural rights come from God, then proof of their existence depends on proof of God’s existence—and further, on proof that God somehow makes rights exist and cannot repeal them. But, again, there is no evidence for the existence of God, much less for the existence of natural moral laws or inalienable rights that somehow emanate from his will.

> To accept the existence of “God” is ultimately to accept it on faith; accordingly, to accept the idea that “rights” somehow “come from God” is to rest one’s case for rights on faith. 

https://www.theobjectivestandard.com/issues/2011-fall/ayn-rand-theory-rights/

cpacia:
I don't think one necessarily needs to take a theist approach to natural rights. There have been a number of libertarian natural rights theorists that did not. 

Personally, I think most people know what is meant by natural law (right to person and property, etc) and including that reinforces the intent of the constitution. 

Although, failure to provide a firm philosophical grounding for natural rights could be a strike against including the phrase. But I would be inclined to leave it.

KacperZajc:
Id say 'law of nature' is the opposite of 'law of god' so this is an atheist approach, I think it can stay but really up to you guys on this one :)

joshmh:
It's true that it's an atheist approach, however it can be argued that the greatest ills of the past century have come from the institution of the secular nation-state. So that may be no great comfort.

My biggest concern is that it really waters down the concept of "explicit consent of the governed," which is a huge departure from "[implicit] consent of the governed [by social contract]." This implicit consent has proved to mean very little. If we postulate that the power claimed in the constitution is authorized either by explicit consent of the governed, or by natural law, those in power can file all their excesses under the rubric of "natural law."

Why do we need "natural law"? The difficulty here is that not all those living in Liberland will consent to be governed, so how may they be dealt with if they intend to do grave harm? That's a very important discussion to have, but I would counsel against sweeping it under the rug of "natural law".

KacperZajc:
#39

PR #10, opened 2015-04-26
define what powers only the PA is allowed to do

tinkyholloway:
The idea being that everything may performed by the PA (if allowed under the Constitution), or by the private sector except the things enumerated in this change.


KacperZajc:
can you elaborate on your statement?

tinkyholloway:
I know this is closed, but let me respond with an elaboration anyhow.

The motivation for this was/is to clearly state two things:

1. there are some things that only the PA is allowed to do. These things are regulated by the PA within the bounds of the constitution. The private sector may not compete with the PA to perform these things/services.

2. Only things that are not enumerated may be performed by the private sector.

For example. The PA might have a monopoly on issuing Warrants, on interpreting the constitution, and being the final arbiter of law, but no monopoly on providing security. The private sector has a number of different securiy companies that provide policing like services. If they want to search someone's house and/or make an arrest, they obtain a warrant from the PA courts.  Perhaps there are a number of competing disputes resolution service providers providing lower level court like services, in this case a PA held monopoly might be final appeal.

I guess it comes down to where you envision Liberland sitting on the Rand-Rothbard spectrum. I personally would like the PA to only do and only be allowed to do the very smallest set of actions/services that are needed to gain international legitmacy and protect our freedoms. Everything else should be provided by the private sector.

PR #11, opened 2015-04-26
clarify that the cabinet is formed from Assembly members

tinkyholloway:
as opposed to anyone that the Assembly deems fit to elect as Chancellor and anyone the Chancellor deems fit to nominate to the Cabinet.


KacperZajc:
I dont think it should be formed from representatives only, it should be formed from experts and those will be very often outside of the assembly

KacperZajc:
no need for change

tinkyholloway:
If Cabinet members are drawn from the citizenry then should this be clarified in the wording? Something like 'from the Citizenry'. Are you proposing non-Citizens as well?

KacperZajc:
we can clarify that Cabinet is formed from citizens;
but not necessarily from Assembly Representatives

tinkyholloway:
No argument from me. I was more interested in the clarification than I was for one way or another.

Issue #8, opened 2015-04-25
Article 3, Section 6


cpacia:
> The criminal offenses shall be prosecuted by the Office of the Public Prosecutor of the Free Republic of Liberland, headed by the Chief Public Prosecutor, on behalf of the Citizens.

I have an objection to this section. Modern governments take a very anti-libertarian position on criminal justice. Just as this section highlights, crimes are treated as crimes against "society" or "all Citizens" instead of whom the crime was actually committed against ― the victim. When criminals are brought to justice, it should not be on behalf of some abstract notion of "society" but rather on behalf of the actual victim and the victim should have the right to direct the prosecution not a public prosecutor. 

Leaving it as is creates the likelihood that injustices will occur. Let me present the case of Aaron Swartz as an example. Swartz committed a crime (albeit a relatively minor one) against MIT. MIT only requested their property be returned and declined to press criminal charges. From the perspective of MIT ― the actual victim ― justice had been served. Case closed. 

Unfortunately, zealous public prosecutors were not content to allow the victim to decide the fate of Aaron Swartz and continued to threaten him with 50 years in prison. Facing a near life sentence, Swartz killed himself. 

Quite frankly the dispute between MIT and Aaron Swartz should have never involved a third party prosecutor. This is an example of the collectivist mentality run amok. "Society" was not the victim and there was no need to pursue charges to get justice for "society". 

I would strike any reference to a public prosecutor from this constitution and treat both criminal and civil cases the same. In both instances, the victim aught to have the right to direct the prosecution as he or she sees fit and should be liable for the defendant's legal expenses if there is no conviction. 

Replace Section 6 with:

> The victim shall have the right to direct the prosecution in criminal cases, so far as is consistent with full respect for the rights of the accused.

naconner:
Very valid point. Though what occurs when the victim is incapacitated or killed? Assume the victim has not entrusted Power of Attorney to another agent.

cpacia:
Good points. I would say n case of death the victim's family or estate would have the right to prosecute. 

There have been some libertarians to suggest this should be a marketable claim subject to homesteading if no one else is available.

For simplicity though you could just have it default to the state as your agent in last resort.

joshmh:
+1

Zephram-C:
What happens when someone kills (or first steals property of, and then kills) familyless person? So there is no one left to indemnify, but basic law was violated. Should this be marketable claim?

PR #12, opened 2015-04-26
require that the PPO receive a complaint

tinkyholloway:
It's not the job of the PPO to go around looking for work to do, instead their work starts with a complaint


cpacia:
+1 this is similar to issue #8.

joshmh:
I agree with the sentiment, but I don't think this change will work as intended. The prosecutor will always be able to get someone to file a claim. So, I prefer https://github.com/liberland/constitution/issues/8.

tinkyholloway:
I agree that this is much better handled in #8

KacperZajc:
there will be offences for which a complaint is necessary to prosecute and some which will be prosecuted automatically - anyway this is not an issue that should be included in the Constitution

KacperZajc:
no need for change

PR #13, opened 2015-04-26
sunset clause for all ordinary bills

tinkyholloway:
has important benefits:
1) it means that we're not saddled with a whole bunch of old laws
2) it keeps the Assembly busy maintaining their existing stack of laws and hindering them from creating more

### DIFF
--- a/Liberland-constitution.md
+++ b/Liberland-constitution.md
@@ -45,7 +45,8 @@ Last changed: Thu Apr 23 2015 12:55:22 GMT+0200 (CEST)
 - **§16.** Any referendum in the Free Republic of Liberland shall be conducted online via the Internet.
 - **§17.** Any Bill vetoed by the Citizens in the referendum shall be considered void and shall not be revived.
 - **§18.** A Bill shall become the law of the Free Republic of Liberland, and shall be known as “Act”, once it is passed by the Assembly in accordance with the provisions of this Article, and either 90 days have passed and no referendum was conducted in accordance with §13 and §14, or on the day the Bill obtained the assent of the Citizens in referendum.
-- **§19.** The law of the Free Republic of Liberland shall enter into force on the day it is signed by the Assembly Speaker, or if unable by another senior officer of the Assembly, which must take place within three working days after a Bill became Act, unless stated otherwise in the Act itself.
+- **§19.** An ordinary Bill passed into law shall expire and become void after a period equal to no longer than two Assembly terms. A Bill so expiring may be reintroduced to the Assembly by the Cabinet per the terms of the Constitution.
+- **§20.** The law of the Free Republic of Liberland shall enter into force on the day it is signed by the Assembly Speaker, or if unable by another senior officer of the Assembly, which must take place within three working days after a Bill became Act, unless stated otherwise in the Act itself.
 
 ## Article II: The Executive Power
 


naconner:
I agree in principle, although another effect would be a higher cost associated with lawmaking, and the unintended repeal of good laws by apathy. Suppose a contentious three percent of the Citizenry (Atricle 1, §13-14), no matter the law, always demanded a referendum after the re-authorization of a law. Basic statutory law would come under a citizen veto every two terms. Not only could these referendums become costly, the citizens could become exhausted from consistently having to vote in favor of criminal codes, tort law, diplomatic bills, etc. With fewer participating, the chances that basic, good laws expire would rise.

cpacia:
I hear what you're saying Neal but I tend to think this would be used to get rid of bad laws far more often the sink good ones.

I suspect if the assembly passes a basic criminal code with no funny business in there.. It probably wouldn't get the 3% needed for a referendum. Almost everyone would support it. It's the other things they might want to criminalize that would cause a problem.

joshmh:
+1

KacperZajc:
cpacia is rights.

KacperZajc:
no need for change

joshmh:
@KacperZajc What do you mean by "no need for change"?

KacperZajc:
no sunset clause is needed in the Constitution

joshmh:
@KacperZajc I believe that most of the people commenting on this thread disagree, include @cpacia who you said was right.

KacperZajc:
sorry I meant naconner

tinkyholloway:
@KacperZajc can you provide some reasons for why you don't think this proposed change is necessary?

KacperZajc:
The idea is to use an extensive Bill of Rights to restrict the areas which the gov can legislate on.
So the majority of laws will be those which are necessary.
Plus, the Cabinet is the only entity capable of proposing a Bill, and that is only once a year upon the State of the Republic speech.
So there are already mechanism which prevent passing unnecessary laws that would have to expire as useless.

rotilho:
+KacperZajc

fare:
There is no reason to believe that good laws that people expect to still be there would be voted down when set for renewal. Renewal should not require a referendum, but still a vote by the legislature and a complete reading, which ensures uncontroversial laws can be renewed quickly whereas bad laws don't last too long.

If you want a law to last long (1) make it short, (2) make it uncontroversial, (3) don't include unrelated articles (as is common practice in the US). Sunsetting is a great filter that allows good laws to survive while killing bad laws.

For more stability, make the sunset longer than one legislature, for instance two or three legislatures.

Note that as far as the threat of instability goes, legislative powers already mean that any law you trust can be repealed, modified, corrupted, etc., by the next legislature. If you really want more stabiity than is inherent in legislation, what you want is not to not have sunset laws, but to have "no amendment" article that prohibits amendments to the law for the next N legislative terms.

KacperZajc:
we have a number of mechanism to limit bad laws:
- limited legislative initiative (only Cabinet and only once a year)
- limited areas of competence of the Assembly
- extensive Bill of Rights 
- referendum veto

I dont think we need the sunset clauses, at least not in the Constitution, it will make the most basic laws unstable.

tinkyholloway:
I agree with @fare. Sunset may promote stability because the Assembly's desire to stop legislation being thrown out keeps the laws that they pass simple, quantity small, non controversial. Nobody is going to allow rape and murder to be thrown out.

I also very much like the combination of sunset and the funding proposal for each bill from @joshmh where each time legislation is introduced its budget is incorporated. I think that is tidy and gets around the problems to tax.

I do like @KacperZajc s "the Cabinet is the only entity capable of proposing a Bill, and that is only once a year upon the State of the Republic speech."

I apologise for being controversial and stubborn. Liberland is my last hope of living free before I die.

cpacia:
I agree with tink. I think we should be doing everything we can to tie the hands of the politicians. 

>limited legislative initiative (only Cabinet and only once a year)
I like this as well, it adds a hurdle to passing legislation, but a hurdle is all it is. I wouldn't be surprised to see the cabinet spending all year writing legislation, propose it at the State of the Republic speech, then have the Assembly spend the rest of the year debating and passing it. It would be unconventional, but thousands of pages could still be passed that way. 

> limited areas of competence of the Assembly
This did not work in the United States. Any honest person who reads the enumerated powers of the Congress would conclude that 98% of the laws passed by the Congress are unconstitutional. Despite that the politicians literally laugh in their constituents faces if they suggest they don't have the authority to pass a certain bill. 

> extensive Bill of Rights
This bill of rights is certainly better than the US bill of rights and hopefully it will do more to limit government than the US bill of rights. Which is why I was so concerned it was being watered down. Without strong language, there isn't much holding back the government. 

> referendum veto
There's no guarantee that people will vote correctly. Voter ignorance and voter irrationality are topics that are widely studied in the public choice field in economics. I would recommend reading economist Bryan Caplan's book [The Myth of the Rational Voter: Why Democracies Choose Bad Policies](http://www.amazon.com/The-Myth-Rational-Voter-Democracies/dp/0691138737).

Which is why I'd like to see as many roadblocks put in place as we can.  A sunset clause is one such roadblock.

KacperZajc:
and now I added:

No Bill proposed to the Assembly shall be passed unless it has been read aloud in entirety by the Assembly Speaker and contain rationale behind it and means of achieving thereof including precise calculations as to its potential impact on the budget. 
+ 
It shall be the duty of any Court of the Free Republic of Liberland to strike down any law as obsolete if there is a significant disparity between the effect of the law and the rationale behind it as declared under §XXX, considering the accuracy of the initial budget calculations and the effectiveness of means adopted.

this responds to your request to make obsolete laws disappear

joshmh:
The sunset clause is far more powerful than asking the court to strike down laws. As @cpacia noted, most federal laws today are obviously unconstitutional except by the tortured reading used by the Supreme Court to avoid having to annul them all. Supreme courts around the world have a terrible track record of actually annulling laws, regardless of how plainly their job is stated in the constitution.

The sunset law allows for much less latitude, and this actually does work in practice. Many laws have expiration dates, and they are either renewed or recognized to expire.

KacperZajc:
and that is why citizens can impeach any Justice of the Supreme Court under this Constitution via referendum

cpacia:
Impeachment power via referendum is likely to only be of use if a justice does things that are grossly out of step with public opinion. 

In practice public opinion is not on the side of liberty and people willfully elect people who will confiscate and redistribute if they fell they will get more in return. 

People who vote to elect people who promise that are not going to vote to impeach politicians or justices who do it.

The point of a constitution should not be to defer to the opinion of the majority of the public but to restrain the majority from violating the rights to life, liberty, and property of the minority.

tinkyholloway:
I find these arguments outweigh the costs. There are two groups of citizens we can't rely on to protect freedom: politicians and voters. The democratic process erodes liberty as shown in the US. We can't leave our most important provisions up to popular vote and we won't get another chance to start over once the looters and narcissists gain a foothold. Given what's at stake I think an overly cautious approach is both warranted and required.

joshmh:
I share the sentiments of @cpacia and @tinkyholloway. We have over 200 years of experience with the US constitution, so we have a good idea of its successes and its failure modes.

I don't see any evidence of impeachment, either by the citizens, or by the legislature, as an effective tool against terrible laws. I feel that the sunset clause could be beneficial and is worth careful consideration.

Another thing to keep in mind is that the constitution can be amended with a sufficient majority. We should err on the side of restricting government because if a restriction is plainly seen as unsuccessful by the public, the government will have every incentive to push through an amendment. If we start by giving the government too much power, it will be very difficult to pry it out of the hands of those holding it.

fare:
This reminds me of this quote:

Our system is defined to prevent majority politicians from overtly oppressing minorities. However, our system is also overtly defined to let majorities elect majority politicians and not let minorities wield electoral power. Kinda funny... one part of our constitution tries to fix a problem which the other half of the constitution tries to create! — Lee Salzman <eihrul@tunes.org>, about the US political system.

KacperZajc:
OK guys if you all insist let's submit the issue to the President to decide. Whatever he decides, we'll accept. I think this is fair. This is a major change and I wouldn't like to introduce it without first consulting the President.

However

If he approves the sunset clauses, I am removing the possibility of striking down legislation as obsolete by the court. To keep both, is to make the law absolutely unreliable.

KacperZajc:
Plus I see a technical obstacle.
If every Assembly (or every other Assembly) has to confirm all laws or otherwise they expire, they will be confirming them as a routine, no one will even bother to read what they are confirming or if forced to read, to analyse whether they are well written - because they cannot change anything in any event.
They wont be able to modify them because they do not have power to initiate legislative procedure.
In 10 years all that will be left from your ideas is a ceremony on the first day of a new term wherein everyone votes 'yes' to confirm all laws in existence. The same way everyone takes an oath now.

fare:
That is why the necessity of reading the law before it's voted combines nicely with the sunset: so that the ceremony becomes costly on the legislators, and they are interested in keeping the book small. That's also why I wanted all the yes votes, not just the Speaker, to read aloud in public the law they are voting for — so they can't claim they have to vote it before they know what's in it.

joshmh:
Vít is doing great work getting Liberland off the ground but it's my strong opinion that in these discussions we should seek to reach a consensus, even if it takes a few extra days. Let's work through the arguments, especially since we have a rather extraordinary group of well-spoken commenters who are volunteering their time for this ambitious project.

History shows that governments excel at writing laws and amassing rules and are much less ambitious about rescinding legislative disasters. We need to apply pressure against bad laws and only allow those with overwhelming consensus.

This sunset clause is one of my favorite ideas, since it's much harder to pass a law than to veto one. Just now, the New Hampshire House and Senate has overwhelmingly passed a law allowing concealed gun carry without a permit, but the governor is threatening to veto it. So she's vetoing the *rescinding* of a law. If there had been a sunset clause on this law requiring a permit, the House would have simply failed to renew it. There would be nothing to veto.

joshmh:
>In 10 years all that will be left from your ideas is a ceremony on the first day of a new term wherein everyone votes 'yes' to confirm all laws in existence. The same way everyone takes an oath now.

Each law would have to be reconfirmed individually, though. So why would a rep vote for a specific law that he was opposed to? With the oath, if he doesn't take it he doesn't get the job. I don't see how these two things are related.

KacperZajc:
my concern is - 
it is safe to assume that there will be no completely useless laws passed in the future (because of restricted areas of competence), only those which will get partly obsolete but considering the fact that the Assembly has no power to modify, they all will have no choice but to vote yes on all of them, because all of them will still be partly useful - they wont veto laws which are at least partly needed. 

Consequently, they will always vote yes in every instance.

fare:
It is NOT safe to assume that no bad laws will be passed. It is NOT safe to assume that an Establishment won't be formed that manages to work around and/or disable restrictions in areas of competence or otherwise — that obviously did NOT work in the US, despite its supreme court.

Assembly has the power to NOT say yes, at which point the bad law expires. Indeed they will vote yes in every instance of a good law — that's how you recognize the good laws. Hopefully, there will mostly good laws in the books — thanks to this and other burdens on lawmaking.

tinkyholloway:
I think the Cabinet should need to reintroduce it. The process followed for the reintroduction should be the same as for a new law. This way it allows them to update it becore the Assembly votes on it. 

It means less irrelevant legal baggage is around making the law simpler

It keeps laws relevant

It keeps laws legitimate since they are either introduced by this Cabinet or reintroduced by this Cabinet, approved by this Assembly, allowed by tnis generation of Citizens

It means controvercial laws have less chance of longevity

It effectivity caps the mass of legislation allowing only the necessary ones to rise to the top because of the burden it places on passing and keeping laws.

Perhaps we could edit this list of benefits and add the downsides so that we have a concise set of arguments to consider.

KacperZajc:
what if the Cabinet reintroduces it but the Assembly refuses to pass?
what happens in the mean time?
we end up without basic laws because part of them was obsolete?

tinkyholloway:
The Cabinet needs to account for the time it takes to move a bill through the process, including the 90 day veto period. For example, if a law were to expire in 2018, then in the 2017 speech to the Assembly where the Cabinet's legislative agenda is outlined, the Cabinet advises the Assembly that it will be reintroducing X, Y, Z laws (perhaps with modification).

KacperZajc:
what if the Assembly refuses to pass the law before the old law expires?

tinkyholloway:
The old law expires.

KacperZajc:
and we are left without the Criminal Code?

fare:
Then the Assembly better vote a renew or code or vote a new criminal code before it expires, if a new code is required. If it's so urgent and all the assembly agrees, of course it will do it.

That said, it's not like you need a criminal code anyway. Between self-defense and restitutive justice, criminals can be dealt with without a criminal code.

KacperZajc:
it amazes me how much confidence you put into the Assembly to do the right thing and vote yes on an urgent law that is due to expire but at the same time you distrust the Assembly in every other matter

joshmh:
I agree with @fare. It's not about doing the right thing, it's about holding onto power. The agents of government generally seek to consolidate and expand their power. They can be trusted to do that. We should worry less about them voluntarily ceding power.

There are already laws in the US that have expiration dates. We see this kind of thing happen all the time. The Public Administration will negotiate with the Assembly to make sure they have a law they can pass. If they're worried, they can submit a weaker backup, too.

tinkyholloway:
Thanks for the hard examples.

I think it depends on the reason for the Assembly refusing to pass. Is there something controversial in the bill? I would think that it would be simple unless a Cabinet at tried a trojan horse. 

Example: Cabinet leaves the Crimes Act until very late, then introduces it with free health care attached hoping that the Assembly will panic and pass.

Worst case situation, the Courts apply common law and the constitution until the constitutional crisis is resolved and the legislation is restored. The application of common law, property right principles and the protections of the constitution would lead a court to the right decision in most cases.

I would expect that the Assembly would then vote to remove the Cabinet or if the Assembly acted improperly, for the Assembly to be voted out at the next election. 

Perhaps the Assembly Speaker monitors the legislative agenda, reminds the house of pending expiring Acts well in advance. If Cabinets don't respond in time to allow fair reintroduction, then the Assembly can tnreaten to remove the Cabinet.

KacperZajc:
@tinkyholloway there will be no common law 'behind' the Acts of Assembly, I want to introduce the principle that all laws are to be found in the Acts so that we wont see the judges expanding their jurisdiction and regulating thing we intentionally left unregulated or worse creating new offence. I think you all agree this is a correct approach.

(There will be elements of common law philosophy and methods but definitely no judges-made laws - this is as dangerous as unrestricted legislator)

I insists we talk to the President and his team on this issue, I will respect his decision. I am not saying this is ridiculous idea but it is 'big' and Id like to run it by him.

tinkyholloway:
@KacperZajc I understand. Perhaps we can formulate the arguments to present. A wiki page?

KacperZajc:
OK go ahead and formulate the arguments and then I'll revise.

tinkyholloway:
I've started a wiki page. I need to step out for 8 hours. Will come back to this and fill the page out then. Thanks for everyone's patience as we work through this. It is clear to me how important Liberland is to everyone here which motivates everyone's desire to see things done well.

KacperZajc:
I am happy to disagree with such knowledgeable people. Thank you.

tinkyholloway:
That's good. It means that whatever the outcome, the discussion will have been hard tested. The Constitution is what will break or make Liberland as a long term haven for peaceful, freedom loving people. All its clauses should be has hard fought and rigorously examined as we're giving this proposal.  It would be easier for anyone to just walk away from their position in the face of opposition. Clearly this matters to people on all sides of the argument.

tinkyholloway:
Wiki page started here: https://github.com/liberland/constitution/wiki/All-ordinary-bills-sunset---arguments-for-and-against

Please help to full this out capturing all for and against arguments. This will allow us to refine, rethink, and refactor before presenting these arguments to a wider audience.

joshmh:
Nice! I'll contribute to this over the next couple of days.

rotilho:
Using development cycle as example, it's clear that stable parts of the systems won't have constants changes and it's not necessary review each parts of the systems every major release.

If the whole system have to be reviewed every major release (election) some stable parts could be lost because of human error causing a possible crash or malfunction of the system.

I agree with @KacperZajc. I can not see this solution working in the real world. I believe sunset clause is not the answer for what you guys are trying to fix.

cpacia:
@KacperZajc the rule is that the cabinet can only introduce laws once per year. What if we allow for laws that don't pass to be amended and re-submitted so you don't have to wait a year to re-introduce a bill that got rejected?

KacperZajc:
I thought of that but it carries a serious risk that they would add some provisions which are completely not related to the original bill and try to re-introduce it, then they can do it non-stop so this is a loophole that gives them power to initiate legislation whenever they want under the disguise of 'amended bill'.

Am I right?

joshmh:
@rotilho, I think a better analogy in the programming world is a cache. If you have a cache where keys don't expire, you'll eventually run out of memory.

cpacia:
@KacperZajc you're right. I thought of that as well. Maybe the provision could require resubmissions to embrace the same subject or else be void.

KacperZajc:
then there will be uncertainty of what constitutes the same subject and how far from the original provisions is too far, I dont think we can pull this off

tbh the more I think about that, the more convinced I am that giving to the judges the power to strike down obsolete legislation on the basis that it no longer matches the rationale behind it is also too much uncertain

we need something in between the possibility to get rid off obsolete legislation and the sunset clauses - it is the same area, the same concern, we dont want to end up with a pile of outdated laws

let's try thinking out of the box here :)

joshmh:
Yeah, good discussion here. I agree that the solution to this isn't in the Judiciary.

In current systems there are lots of incentives to add laws and hardly any to rescind laws. So the most obvious thing to do is to make it easier to rescind laws. For instance, give the Assembly the power to rescind any law at any time with 1/4 of the votes (the exact number can be debated). It's already quite difficult to pass a law. Also, maybe allow some kind of referendum to rescind a law.

The problem with this is that reps are dumb and lazy, and they're not good at hunting down obscure laws that don't have to be there, even if it's relatively easy. That's where the sunset clause comes in, which flips it around by making them work harder if they want to keep it.

A law is valid for two terms (8 years). However, the Cabinet can always ask the Assembly to reaffirm a law, which gives it another 8 years. This has to be so, because otherwise the Cabinet could just ask the Assembly to pass a new law that annuls the current law and adds it again as a brand new law. Even if this wasn't allowed, they could just make cosmetic changes and pass it as a new law.

If a law is super important, I could see the Cabinet and Assembly reaffirming the law every year, so they'd never get close to worrying that it would expire. Even if the public or the Assembly vetoes the reaffirmation, the original law would still stand, but would expire on schedule. And the next year, the cabinet could attempt to reaffirm it again.

So maybe we don't have to worry about important laws expiring.

KacperZajc:
All laws passed by the Assembly under the Constitution shall expire within 8 years unless the Cabinet requests the Assembly to reaffirm the law in its entirety, which the Assembly must do.

If no request is submitted to the Assembly, the Assembly can only confirm the law with the majority of no fewer than 15 votes.

does it makes sense or it is toothless?

joshmh:
What's the point of having the Assembly reaffirm the law, if they have no choice? The problem is that the Cabinet is the executive branch, so they will have the most incentives to keep bad laws. I think we're getting there, though, it's just a matter of hashing out the exact number of votes needed to pass, reaffirm and rescind.

@cpacia's notion is that there's should be a 3/4 majority in favor of a law at all times. If at any time this isn't true, they can rescind with 1/4 majority. If a law comes around for reaffirmation, you'd need a 3/4 majority in favor as well. Are the arguments in favor of relaxing this 3/4 majority for reaffirmation and rescission?

KacperZajc:
All laws passed by the Assembly under the Constitution shall expire within 10 years.
The Assembly can reaffirm the law by a majority of no fewer than 15 votes.
If the law is not reaffirmed, the Cabinet shall have the power to initiate legislation on that very same subject independently of the power to initiate legislation upon the Sate of the Republic speech.
Such Bill shall be passed with no fewer than 15 votes.
If no Bill is passed, the Cabinet shall not submit any proposal more proposal on the subject outside the framework of ordinary power to initiate legislation. 

???

rotilho:
Are there any way to keep stable laws? eg, after the first review (10 years) there is no need for further review

rotilho:
Thinkg out of box:

- Every law must go through a trial period until be reaffirmed.
- Every reaffirmed law must go through a extended second trial period until be reaffirmed and become stable

joshmh:
@rotilho I disagree that a law can ever be stable. Circumstances, moods and norms change over time. Sometimes it takes decades for the public to realize how destructive a law is. For example, consider the change in public opinion about marijuana prohibition over the last few decades.

KacperZajc:
could you all comment on my last version of the clause?

tinkyholloway:
@KacperZajc  I think it's looking good. I just have one question:

Why do you prefer to have the Assembly vote directly on sunsetting laws and not have the Cabinet reintroduce. It introduces more exceptions to the normal law introduction process so I'd like to understand the benefits.

If it is to allow the Cabinet the right to edit a Bill that has failed once in Assembly, then I wonder whether this single retry might not be warranted on all Bills.

My motivation here is only to keep things simple and consistent to avoid giving lawyers exceptions to push and play.

joshmh:
Almost there.

> All laws passed by the Assembly under the Constitution shall expire within 10 years.

I don't think the exact number is crucial here, but what's the rationale behind extending it to 10 years?

Other than that, I agree with @tinkyholloway. I don't see any benefit in a special case here. Reaffirming is just introducing a law saying "law x is passed anew." It doesn't have to be a special case. We don't need to give the Cabinet the power to call a special session to reaffirm a law, since they had 10 years to extend this law and failed.

tinkyholloway:
@rotilho I understand what you mean by stability, but I think this is better achieved by encouraging Bill authors to write things that are noncontroversial. There is no benefit to kilograms of useless, stable legislation. It works against the principles of the population understanding the law (because of the volume) and the laws being relevant (because of outdated content).

How much sooner might slavery have been abolished had the slave codes expired because parliament did not have the stomach to renew. 

Stability can become a tyranny. Based on history, I'd say that it probably will.

KacperZajc:
maybe instead of a sunset clause we empower the citizens to rescind any law whatsoever in a referendum triggered by a request of 3% of population any time?

joshmh:
I like the idea of allowing the citizens to rescind by referendum, but that doesn't replace the sunset clause. We can't expect the citizens to be vigilant about removing all the bad laws from ten years ago. Also, if we want to explore funding laws for their entire term, we need the sunset clause.

KacperZajc:
let's get rid of the judiciary power to strike down the laws as obsolete and do not introduce this referendum idea but try to introduce sunset clauses IF the president agrees

we need to finish the WikiPage on that and propose sensible wording

joshmh:
:+1:

tinkyholloway:
@KacperZajc I'd be interested in hearing - should you so oblige - as to whether you've moved from your original position on this or whether you're simply working through a process. If we can't convince you being someone deeply involved in the argument, then I fear this will be a difficult challenge when we present the case in one shot to someone.

Either way, thank you for sticking with this. I'd be hugely grateful if you could mount a strong anti position on the wiki to allow a counter anti argument to be presented. Given that most people reading the proposal will be initially against it - this proposal exists nowhere else - these counter anti arguments will be our only chance of a 'discussion' like we've had in this thread.

tinkyholloway:
@joshmh your funding idea is awesome. Let's make that a highlight in the wiki write up. It and this really could make things simple and tidy.

KacperZajc:
I still oppose the idea but I want to give it a fair hearing.
And I want to leave this decision to the President.

manchfu:
I'll just add that to me @tinkyholloway sunset clause + @joshmh 's direct funding idea together are the most exciting parts of this constitution. Everything taken from the US constitution sounds great, but the US constitution is pretty worthless to the US these days.

KacperZajc:
I dont see how we can provide for direct funding. it unenforceable and amounts to a direct democracy.

But I have an idea as to the sunset clauses - 
we could add a provision that the requirement of sunset clauses kicks in after the Constitution has been in force for 10 years - 
by this requirement we give ourselves time to make the most basic laws (which will be essential thereby should not expire), we are secured because it is obvious that it will be us who will control the Assembly for the first years and we make sure that the laws produced by the subsequent governments (which will not be so essential) expire every 8 years.

This seems to me like a fair offer.

manchfu:
As I understand it, it would be instead of a budget, so there's no other way for a bill to be funded. It's essentially crowd-funding for laws, only they have to pass the ordinary democratic channels as well.

A bill that is voted in is either funded within a period of time and becomes law, or void. It also forces government to be frugal and be on budget.

joshmh:
> I dont see how we can provide for direct funding. it unenforceable and amounts to a direct democracy.

The money could be collected up front, at the time of voting. Or, it would be considered a contractual obligation of each voter, since they explicitly agreed to it. The government could try to collect via the usual court system under contract law. If not enough money is collected, the fund for the law runs out and no more invoices can be issued for that law.

It's not direct democracy because the formulation of the laws is done by the Cabinet and Assembly. The citizens are paying for the laws, but that's true in every government.

KacperZajc:
and if they wont pay for the courts we will end up with no administration of justice. sorry guys but I really cant imagine putting such a provision in the Constitution...
it will not be a state it will be a lawless anarchy if we adopt all your ideas here

joshmh:
Didn't Vít already say taxes would be voluntary? Is the Cabinet going to go around Liberland locking up anyone who doesn't fund them?

manchfu:
If nobody pays taxes you won't have any money to do anything about it, and I think we all want taxation to be voluntary on some level. Does anyone here think people should be incarcerated for not paying the government?

KacperZajc:
but normally you'll have one vault and you could move money around to finance the most basic functions from voluntary donations, under your proposal if citizens say no to the criminal justice act then you wont be able to finance it 
you are building a utopia here, it will never work even if we assume that you can finance the whole state from donations

KacperZajc:
or what if they agree but then wont pay? or move away from Liberland?

manchfu:
I think so far the issues you brought up have been with the details, not necessarily the idea. If at all, the number of deceased or other non-payers, will be quite low, as with Kickstarter. 

There's a Bitcoin platform as well, Lighthouse - https://www.vinumeris.com/lighthouse

If there are 10,000 people living in Liberland, or even 30,000 - the justice system should be small and very low budget. What do you really need other than office space (preferably rented) and some salaries?

I also don't think it's intellectually honest to use the "what if nobody pays" argument without asking "what if nobody votes" for a judicial system. What if the assembly unanimously vote that there should be no judicial system? This is a technical possibility, and similar to direct funding, improbable.

KacperZajc:
insofar as sunset clauses might make sense because as someone pointed out the government has a strong incentive to renew the laws because they give it powers,
there is no direct incentive for the citizens to pay for anything.

maybe it will be low budget but it needs to be stable and a sure thing, you cannot have a state in which the whole judicial system might stop exist next day

this is a ridiculous idea, beyond my imagination :(

tinkyholloway:
I think there are only two ways to limit the growth of the State:

1. increase the burden/difficulty of introducing legislation. It has to be a real burden and not just some pretty words that will be debated forever in the background while the State marches on. Some effort was made with the US constitution in this regard, but it is obviously a failure. This method alone cannot limit the power and growth of the State.

2. provide limits on how the State can raise revenue. The Constitution provides some good protections here already: a) not allowed to issue currency; b) not allowed to regulate currency; c) not allowed to run deficit budgets; d) not allowed to obtain revenue from anywhere that is not specified in the Finance Bill.

Given that taxation is a direct threat to the right to private property where the State can simply demand payment (single fee, flat rate, progressive?), this needs to be cleared up in the Constitution. I have seen statements that taxation would be voluntary but none of these have made it into a draft yet.

If taxation were voluntary, then surely direct funding of bills leads to justice being continually funded given that the population consider it important. If paying for justice comes with a lot of controversial baggage and payment of taxes is voluntary, then I would see that as a threat to capturing that revenue. Direct payment means that obvious things get easily funded and controversial things perhaps don't. We might even see insurance companies contributing towards justice since it would be in their better interest to do so.

I think the Constitution needs to be bold. Creating a Constitution that is the US constitution with some good, but minor changes is likely to lead to a State like the US with some minor changes.

If this turns out to be a disaster, then there is a Amendment clause. It will be easier to Amend towards socialism than it will be to Amend towards freedom.

KacperZajc:
To clarify what you said:

There is a mandatory referendum for any Bill that would introduce taxation. This was expressly approved by the President in my conversation with him. This safeguard amounts to voluntary taxation.
What is more, there is prohibition of income taxes in any event.

I have also 2 more points to make apart from those I've already expressed:

1) Direct funding of every Bill will lead to corporations financing Bills they like. the situation in which corporations have a power to influence the government (and possibility to finance a bill is a power to influence the government) is even more dangerous than limitless government itself. I dont think I need to convince anyone here about that. They will be striking deals with government so that is passes bills that will get their funding but at the same time, guess what? - they will be bad for the free market and average citizens. You idea will lead to corporatism. 

Alternatively: 

2) I know you all assume we can do whatever we want but this is not the case. If we want to be recognised by others states and we really want, we have to look like a legitimate state. Your idea might result in anarchy, not a limited government. We wont be able to join any international organisation and again we want, nor will be treated seriously by our neighbours. 

And again:

If there is one think we need is a stable and efficient system of justice and I dont want to run the risk of loosing it tomorrow because we wont get funding.

Finally:

The power to finance or refuse to finance a bill is nothing more than introducing direct democracy through the back door. This was something I wanted to avoid from the beginning. We have the referendum option and this is enough. If they dont want that bill they can veto it. 

To sum-up:

This is a really BAD idea guys and I can't agree to it. Sorry.

tinkyholloway:
> There is a mandatory referendum for any Bill that would introduce taxation. 
> This was expressly approved by the President in my conversation with him. 
> This safeguard amounts to voluntary taxation.

I don't understand @KacperZajc can you confirm, whether the view held is that the majority voting to authorise a tax, is what is considered "voluntary"?

KacperZajc:
This actually sounds misleading, sorry.
It is voluntary in a sense that the budget is supposed to be based on voluntary donations.
There is absolute prohibition on income taxes, however, if the Assembly would like to levy other forms of taxes there is a mandatory referendum requirement - so it is also 'voluntary' in a sense that it can be blocked by the Citizens.

This formula was accepted by the President because it is hard to estimate the overall size of potential donations and send a signal to our international partners that we are not dead against any forms of taxes (even if we are) which would make them even more hostile towards us.

FellowTraveler:
+1

The number of U.S. laws is so large that a sunset clause is clearly necessary.

Quote: http://www.kowal.com/?q=How-Many-Federal-Laws-Are-There%3F

No one knows how many laws there are in the United States.  Apparently, no one can count that high.

They’ve been accumulating, of course, for more than 200 years.  When federal laws were first codified in 1927, they fit into a single volume.  By the 1980s, there were 50 volumes of more than 23,000 pages.

And today?  Online sources say that no one knows.  The Internal Revenue Code alone, first codified in 1874, contains more than 3.4 million words and, if printed 60 lines to the page, is more than 7,500 pages long.  There are about 20,000 laws just governing the use and ownership of guns.

New laws mean new crimes.  From the start of 2000 through 2007, Congress had created at least 452 new crimes, so that at that time the total number of Federal crimes exceeded 4,450.

FellowTraveler:
Maybe should automatically sunset laws after 20 or 30 years. After all, the world belongs solely to the living -- no generation of the dead has the right to bind those yet unborn. That way the obsolete laws will pass away automatically, and the good ones will be kept.

It does seem a little ridiculous to have to renew laws every single session of the Assembly (every single year?) but 20-30 years is a lot more reasonable.

cpacia:
Good laws... sure a 20-30 year time span is plenty. But if bad laws are passed I would want them to have to vote on it again much more frequently. 

Granted in the US this doesn't really as the patriot act expires frequently and they re-up it every time, but at least they have to regularly go on the recording voting for it.

FellowTraveler:
> But if bad laws are passed I would want them to have to vote on it again much more frequently.

Yeah but in Liberland, all laws are subject to a Citizens' veto before they go into effect. That seems good enough to me to give it a 20 year sunset period, no?

fare:
In twenty years, the communists have ample time to invade the schools and explain to every kid how their oppressive laws are "normal". A five year period is a better compromise between not having to renew too often and having to renew regularly. Also — it allows to tie the law to its budget, and avoid the budgetting prestidigitation typical of western democracies.

Jean-LouisMesic:
Bumping @KacperZajc last comment on the sunset clause.

Setting up the good legislative infrastructure of Liberland before implementing the sunset clause may possibly be the best way to go about this issue.

Thoughts?

Auahi:
I'm very late to the conversation and sorry if I'm bringing up old points but won't this make the law making job next to impossible, Imagine the work load in two years from the time the constitution is put into place!

Also this allows for basic laws like free speech to be easily questioned in the future, maybe the sunset clause could only apply to select laws...

PR #14, opened 2015-04-26
Jury power

tinkyholloway:
truly being judged by one's peers

### DIFF
--- a/Liberland-constitution.md
+++ b/Liberland-constitution.md
@@ -77,7 +77,7 @@ Last changed: Thu Apr 23 2015 12:55:22 GMT+0200 (CEST)
 - **§4.** The Supreme Court shall be presided over by the Chief Justice of the Supreme Court, and shall be composed of him or her and four others judges called the Justices of the Supreme Court, appointed by the Chancellor for life, and shall reach its decisions by a majority of votes.
 - **§5.** The Supreme Court shall hear appeals from the Civil Court and the Criminal Court on the point of law where such appeal was permitted by the Supreme Court.
 - **§6.** The criminal offenses shall be prosecuted by the Office of the Public Prosecutor of the Free Republic of Liberland, headed by the Chief Public Prosecutor, on behalf of the Citizens.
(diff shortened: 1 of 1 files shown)


KacperZajc:
this is not how it works
Jury doesnt interpret the law because they have no idea how
Jury only decides matters of facts, which are true which are not
what you are proposing is impossible

naconner:
Juries have a long history of juding the the law they're being asked to apply:
http://en.wikipedia.org/wiki/Jury_nullification

A jury's right to acquit, regardless of guilt, is the people's last check on immoral laws or a corrupt system.

KacperZajc:
sure and it can acquit under the current wording of the paragraph as well, but to put what was proposed here would be to distort the whole system

KacperZajc:
no need for change :)

cpacia:
Apparently Thomas Jefferson was one of Vit's inspirations. He also had this to say,  "I consider trial by jury as the only anchor ever yet imagined by man, by which a government can be held to the principles of its constitution."

It was envisioned by most of America's founders that the Jury would be the final defense against an out of control government.

JOHN ADAMS (1771): It's not only ....(the juror's) right, but his duty, in that case, to find the verdict according to his own best understanding, judgement, and conscience, though in direct opposition to the direction of the court.

It has devolved to the point where the government's own Judge interprets the law exclusively, prevents the defense from raising questions about the law itself, and throws jurors off the case if the government believes they will not follow the judge's instructions. 

That is, and has been, a recipe for tyranny. Any constitution worth its salt must sanction jury nullification.

And if the jurors are not capable of interpreting the constitution, then why do we have one at all?

KacperZajc:
it is the judge which directs the jury how to interpret the law 
HOWEVER
the jury can refuse to convict under this interpretation anyway even though they should and this is what jury nullification is all about - this is what is written now

tinkyholloway:
If the jury reaches a decision of nullification can the prosecution appeal or is it trial over?

KacperZajc:
no, never, that is the rule, but we can clarify that if you want :)

KacperZajc:
#

PR #23, opened 2015-04-27
Restore Bill of Rights §6 and §7 to their original language.

cpacia:
These two lines were said to be too restrictive, but that was intended to be the point.

If the purpose of government is to protect person and property, then those two lines do exactly that.

Every consitution that has ever been written has failed to limit government. If historical precendent holds, this one will fail as well. But if we make the language as restrictive as we possibly can then maybe we stand a change.

Watering down the language to make it easier for the government to create laws will just expidite the erosion of the constitution.

### DIFF
--- a/Liberland-constitution.md
+++ b/Liberland-constitution.md
@@ -95,21 +95,22 @@ The Bill of Rights shall constitute the integral part of the Constitution and sh
 
 * **§4.** No ex post facto law shall be passed.
 * **§5.** No Bill of Attainder shall be passed.
-* **§6.** No law shall interfere with voluntary consensual and/or contractual relations among adults, or the right to form cooperative ventures of any kind so long as no rights of other persons are endangered. 
-* **§7.** No law shall invade the privacy of peaceful persons by unreasonable searches, seizures, storage of personal information without consent of their rightful owners, including but not limited only to the use of digital means, freeze of assets of any natural and/or legal person, otherwise than in pursuit of a Warrant issued by a judge of appropriate court, and only to an extent which is necessary to attain the end; no Warrants shall be issued, but upon probable cause, supported by appropriate order, and particularly describing the place to be searched, and the persons or things to be seized or put under surveillance.
-* **§8.** No law shall allow for the private property be fully or partially taken for the public use without the consent of and/or mutually agreeable compensation to the owner.
-* **§9.** No law shall create victimless and/or consensual crimes.
-* **§10.** No law shall mandate any person to associate and/or transact with any other person; nor shall it prohibit it.
-* **§11.** No law shall abridge the freedom of thought and feeling, or their peaceful expression or dissemination, as in speech, press and other media, artistic depiction, or religious practice; nor shall any law promote or hinder any religion, artistic culture, gender or specific community.
-* **§12.** No law shall abridge the right to assemble peacefully where no rights of others are infringed. 
-* **§13.** No law shall establish the institution of slavery, conscription, indenture, or any other form of involuntary servitude other than the Jury duty, within the Free Republic of Liberland, or in any place subject to its jurisdiction.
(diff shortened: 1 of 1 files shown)


KacperZajc:
as I said

I did this before I was showed GitHub and here is why:
This line sounds great, I know, but it is too broad, really too broad -
- it might be later used to prevent any sorts of regulations we might need for example due to international agreements we will need to sign;
- it would prevent for example mandatory vaccination (Im not saying we want one, I dont know what is the president's stance of this?);
- it would allowed full abortion up to birth (again I dont know if we want to allow, I dont know what is the president's stance of this?);
- it might be used to block fines of any sort;
- it might be used to to block spacial planning of the territory of Liberland (that we definitely need because we have only 7km2 what President confirmed)

cpacia:
Josh this pull just adds back these two lines:

**§6.** No law shall abridge the right of each person to do as he or she chooses with his or her own person and property, so long as he or she does not interfere, by force or fraud, or the threat thereof, with the equal right of others to do as they choose with their own persons and property.
+* **§7.** No law shall abridge the right of persons to the control of their own bodies, nor interfere with voluntary consensual or contractual relations among adult persons, or the right to form cooperative ventures of any kind. 

The rest of the pull is re-numbering.

KacperZajc:
I wanted to get rid of §6 and 'the right of persons to the control of their own bodies' part of §7 for reasons I explained

naconner:
By removing the language about a person's right to control their own bodies, that also clears the way for drug prohibition.

One of the philosophical precepts of liberty is an individual's right to self-ownership. Is this not one of the pillars of Liberland too?

> it might be used to to block spacial planning of the territory of Liberland (that we definitely need because we have only 7km2 what President confirmed)

Zoning in Liberland? Eminent domain? I see this was changed too:

Prior:
> No law shall allow for the private property be fully or partially taken for the public use without the consent of, and mutually agreeable compensation to, the owner.

Now:
> No law shall allow for the private property be fully or partially taken for the public use without the consent of **and/or** mutually agreeable compensation to the owner.

How is this consistent with what's written on the official website?

> Liberland currently needs people who ... have respect for private ownership which is untouchable.

KacperZajc:
zoning was mentioned by the President

KacperZajc:
anyway we can change it back guys if you all insist I am not going to defend these proposed changes, I made them to enable singing international agreements and spacial planning but I dont want you to consider me as pro-government

joshmh:
@KacperZajc We're all entitled to our considered opinions. I think everyone here just wants a carefully thought through discussion of this constitution. As someone living in the US, I'm intimately familiar with the ways those in power can bypass even the most noble sounding working, and I think that's what some of the commenters are trying to express here.

KacperZajc:
what would you say about a combined provision:

No law shall abridge the right of persons to the control of their own bodies and property, nor interfere with voluntary consensual or contractual relations among adult persons, or the right to form cooperative ventures of any kind so long as no rights of others are interfered.

KacperZajc:
???

manchfu:
I feel this is off to a bad start as far as limiting the power of government, so any change someone makes taking away power from the people should be scrutinized. I don't think we have to rush this either.

If Vit thinks any provision is necessary, he should join the discussion that we're having thanks to him.

tinkyholloway:
mandatory vaccination and/or late term abortion might be covered in the original wording by 'not interfere with the rights of others'.

Zoning can be covered by considering that all land is 'currently owned' by Liberland Company. When sold a clause can be added to the sales agreement limiting use.

International agreements can be covered by the treaty section in A1.

Fines in the part where it talks about restitution. Unless you're talking about the state punishing people in which case I think you clarifying this would be helpful.

KacperZajc:
@tinkyholloway 
if we restore §7 abortion and vaccines wont be covered 
I agree with your itnerpretation if we merge §6 and §7 into: 

No law shall abridge the right of persons to the control of their own bodies and property, nor interfere with voluntary consensual or contractual relations among adult persons, or the right to form cooperative ventures of any kind so long as no rights of others are interfered.

PR #24, opened 2015-04-27
Add back self-defense

joshmh:
This was just removed, so I'm adding back in so we can discuss.


KacperZajc:
this was removed on the personal request of President's assistant who is a criminologist and said it would be better to keep the right to bear arms separately from the right to self-defence which should be left to the jury.

joshmh:
@KacperZajc I left the sections separated, but this lays out explicitly that there is a right to self defense and that one may defend himself with force against unlawful actions of agents of the state. I feel this is a cornerstone of the consensus we're trying to establish, and is certainly not self-evident based on my familiarity with the US and Israel. (Israel has a similar provision to what I added back, the US does not, except in a few states.)

KacperZajc:
we can add this then:

The right to self-defense against initiators of aggression, including the agents of the Public Administration where acting unlawfully, is guaranteed

KacperZajc:
oh sorry I havent noticed, yours is actually good :)

joshmh:
:+1:

KacperZajc:
add this Right but in the Individual Rights section of the Bill of Rights

Issue #27, opened 2015-04-27
Every law needs include an explicit rationale


fare:
And citizens can sue and nullify the law if its rationale fails to hold, is not valid anymore, if the desired effect isn't achieved by the law, etc.

This makes it harder for laws to be remain on the book when they utterly fail at what they claim to do and are actually a way to subsidize special interests through their "unintended consequences".

fare:
(This issue is split from #20)

cpacia:
+1

joshmh:
+1

tinkyholloway:
If sunset #25 #13 were required, this would be covered, no? If not can you help to explain the difference.

KacperZajc:
#28

PR #28, opened 2015-04-28
Every law requires a rationale

fare:
See issue #27.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,21 +4,15 @@
 
 **Notice: this is NOT a final version of Constitution, it is just a draft!**
 
-- [Free Republic of Liberland Constitution draft](Liberland-constitution.md)
+* [Free Republic of Liberland Constitution draft](Liberland-constitution.md)
 
-### Generated formats
+## Maintainers
+
+* Kasper Zajac - [@kasperzajc](https://github.com/KacperZajc)
+
+## Generated formats
 
 [Markdown], [PDF], [HTML], [ePub], [Mobi]
-
-### Translations
-
-Do you want translate the constitution into your language? Please continue to [liberland/translation](https://github.com/liberland/translations) repository.
-
-You can view and download all languages in [liberland/docs](https://github.com/liberland/docs) repository.
-
-## About Liberland
-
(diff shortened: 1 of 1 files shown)


KacperZajc:
a bit too long plus add the requirement to read it aloud:

No Bill proposed to the Assembly shall be passed unless it has been read aloud by the Assembly Speaker and includes rationale and means of achieving thereof which does not conflict with the Constitution.

fare:
I'd make the reading aloud a separate provision, and I would insist that no one may vote positively to introduce, amend or keep from expiring a law unless he has publicly read its entirety aloud during the legislature.

Note that all sponsors may read it aloud at the same time while the bill is being introduced for a discussion or vote, but that needs not be mentioned in the constitution itself.

KacperZajc:
this provision is too long really, and I would combine those, there is no reason not to; here is what you want in a short form

No Bill proposed to the Assembly shall be passed unless it has been read aloud in entirety by the Assembly Speaker and includes rationale and means of achieving thereof which does not conflict with the Constitution.

fare:
Actually I would include a budget restriction with every law, which plays well with nullification for bad cost/benefit analysis, and also plays well with sunsetting: the budget has to be renewed anyway. Linking the budget to individual laws also prevents the system of graft, opaqueness and log-rolling that appears when you vote budget as a single law.

That's a much better thing to put in the same article as the "must be read" provision, that is better to combine with sunsetting, since the two play together into limiting the overall size of the legislative code.

KacperZajc:
we can add the requirement of budget but I dont agree with sun-setting as I wrote in various places 

No Bill proposed to the Assembly shall be passed unless it has been read aloud in entirety by the Assembly Speaker and includes precise calculation as to its potential impact on the budget as well as rationale behind it and means of achieving thereof, which does not conflict with the Constitution.

???

KacperZajc:
we need stable laws, we need law that people now, we need to be predictable, how will you teach lawyers what the law is if it change change entirely every 4 years? not to mention that normal people dont have time to track all laws all the time because they all can change at once

i understand your intention guys but this is not the way to go

Issue #17, opened 2015-04-26
Consider using Markdown numbered lists for sections


joshmh:
Using Markdown numbered lists would make it much easier to add and remove sections without needing to renumber everything. It would also dramatically reduce noise in the visual diff. References would be an issue, but I think it's worth the tradeoff.

Example:

```
## Article II: The Executive Power

1. The executive power of the Free Republic of Liberland shall be vested in the Cabinet.
1. The Cabinet shall be formed and presided over by the Chancellor, who is the Head of State.
1. The Chancellor, on behalf of the Cabinet, shall once a year publicly give to the Assembly information on the State of the Republic, and recommend to their consideration such measures as the Cabinet shall judge necessary and expedient thereby initiating a legislative procedure in accordance with the provisions of Article I.
1. The powers of the Chancellor and the Cabinet shall be restricted to the following only:
  1. to run day-to-day state affairs;
  1. to initiate a legislative procedure including the preparation of the Financial Bill containing the Annual Budget without a budget deficit;
  1. to be Commander in Chief of the military, once the state of war is declared by the Assembly;
  1. to protect and promote the interests of the Free Republic of Liberland worldwide;
  1. to make treaties with the consent of no fewer than nineteen votes of the Assembly Representatives;
  1. to execute in good faith the laws passed by the Assembly;
  1. to produce byelaws in a form of Delegated Instruments under the powers expressly granted to it by Acts of the Assembly.
1. The Cabinet shall consist of the following Departments only:
  1. Department of Justice
  1. Department of Internal Affairs
  1. Department of Security
  1. Department of Foreign Affairs
  1. Department of Treasury
1. Each Department shall be headed by the Secretary for the respective Department who shall answer to the Chancellor and the Assembly.
1. It shall be the duty of the Cabinet to refuse to execute any laws passed by the Assembly in conflict with the Constitution.
```

rotilho:
+1

naconner:
Sure beats quills and hemp paper. :+1:

rotilho:
I changed my mind. It would be a problem when a item refers another one.

joshmh:
Yeah, I agree that references are a problem. Since I think improving PRs will have a huge effect on everyone's participation, I'm going to keep stabbing at this.

Perhaps we could try a simpler approach:

All new additions in a pull request should be written as the last item in the section, regardless of whether it would make more topical sense to group it with items higher up. We can then all discuss the PR easily.

[...]

PR #42, opened 2015-04-28
Contributing

joshmh:
Added CONTRIBUTING.md as per https://github.com/liberland/constitution/issues/17


CONTRIBUTING.md
```diff
@@ -0,0 +1,27 @@
...
+
+If you want to fix a series of typos, you can submit a pull request that incorporates all of your typo changes in the document.
+
+## Adding a line
+
+If you want to add a new, numbered line to the constitution, add it as the last item of the relevant section. This way you won't have to re-number all the other items, which would make it very hard to review the change. Don't worry if the change would make more sense next to a higher-up item. That will be fixed after all the changes are discussed.
+
+## Removing a line
+
+If you'd like to remove a line, delete it but don't update the numbering of any of the other lines. This will be fixed in the end. For now, we want to keep your change restricted to the line you're changing. Make sure you fix any references to the item you deleted.
```
hazzik: I would like to say something like "$II.100 Is obsolete now" instead of just removing the line

joshmh: The diff will show the deletion. You can write that in the commit comment.

burningtree:
I do not think it is necessary to guide all to use the command line. These repositories will contain only text files, so everything can be easily edited using GitHub web interface - this is not a software project.

We need help from the public, it is pointless to send lawyers to the command line.

joshmh:
Oh cool, I didn't realize that creates a pull request. I'll put that in instead.

burningtree:
Two files are unnecessary, please consolidate it into one.

CONTRIBUTING.md
hazzik: The diff - yes, but the numering itself will lead to confusion. So, people just have to see that this point is no longer valid.

joshmh: Ok, that makes sense, I'll add it.

PR #43, opened 2015-04-29
Add back "force or fraud"

joshmh:
Yes, I know this was raised in https://github.com/liberland/constitution/pull/23 and the PR was closed. I'm proposing this again because I don't believe consensus was reached and because this is a crucial part of the document.

I'll summarize my view of the debate here. Both sides agree that there is a very substantial difference in what laws may be passed between the two versions.

Those against the change posit that this proposed limitation restricts the government too much and would prevent laws establishing zoning, mandatory vaccines, abortion laws, fines and international agreements.

Those in favor of the change believe that the current language means to represent something like the Non Aggression Principle, but in fact is toothless. Evidence for this is that similar language already exists in some constitutions and has not stopped the litany of laws in stark opposition to the principle of non-aggression. They also maintain that the proposed laws are either undesirable or would be constitutional even after the proposed change.

This is a weighty subject. Let's give it a few more days of discussion.


KacperZajc:
maybe we put it like this:

No law shall abridge the right of persons to the control of their own bodies and property, nor interfere with voluntary consensual or contractual relations among adult persons, nor impinge on the right to form cooperative ventures of any kind, so long as no rights of others are infringed without their consent or unbeknown to them.

would this potentially disallow abortion and mandate vaccines?
would this be sufficiently sharp?

joshmh:
I feel like we're running up against some of the problems with the Non-Aggression Principle. This frustrates the goal of completing this document, but on the other hand it's pretty cool that we're in the business of implementing some of this new thinking. A great deal of thought and debate went into the US Constitution, as well.

I don't have a good answer for this, yet. It all goes back to the futility of trying to control future powerful people with a few strokes of the pen. That's why I favor thinking carefully about the structure of power we're specifying. Let's keep this open a bit, and see if we get any more discussion.

KacperZajc:
yes this is a problem with the application of the non-aggression principle
could you analyse the wording I proposed?
am I correct in my assumptions as to its potential effects?

KacperZajc:
do you guys accept this version?:

No law shall abridge the right of persons to the control of their own bodies and property, nor interfere with voluntary consensual or contractual relations among adult persons, nor impinge on the right to form cooperative ventures of any kind, so long as no rights of others are infringed without their consent or unbeknown to them.

joshmh:
I can't say that this is acceptable to me. This makes everything forbidden. Everything today is considered some kind of right. Most of Europe considers free healthcare a right. So if I don't pay for someone's free healthcare, can the state interfere?

KacperZajc:
but you didnt object the 'interfere with rights parts', you just wanted to add 'by force or fraud'
this part must stay, this is the essence of the non-aggression principle, this is what we believe in here isnt it?

joshmh:
The "force or fraud" part limits what we mean by "infringing on rights". The NAP says you can do whatever you want as long as you don't *aggress* against another, which means force or fraud. This is saying you can do whatever you want as long as you don't infringe on another's rights, which could mean anything.

KacperZajc:
to aggress against others is to infringe upon their rights, what other explanation of aggression do you know?

joshmh:
Yes, the troublesome part is the infringing upon others' rights that's *not* aggression. If we're only protecting against aggression, we can write something like:

> No law shall abridge the right of persons to the control of their own bodies and property, nor interfere with voluntary consensual or contractual relations among adult persons, nor impinge on the right to form cooperative ventures of any kind, except insofar as exercising such a right aggresses against others.

KacperZajc:
No law shall abridge the right ... so long as no rights of others are infringed by force or fraud, or threat thereof, or unbeknown to them.

???

Zephram-C:
This is a difficult subject. Some extreme cases: 
If we accept *right to health* then wearing scary (or offensive) T-shirt can be *infringing on others mental health* or even *infringing on others mental health unbeknown to them*. 
If we stay with force or fraud only, then unintentionally infecting someone with bacteria or viruses by simple contact (due to not vaccinating oneself) is what? Is this application of force same as intentionally infecting oneself and then intentionally infecting others? Or is it *(not vaccinating oneself)* fraud, (because others can implicitly reasonably expect everyone silent being vaccinated, HIV-negative, etc.) ?

joshmh:
Yep, this is the core of the debate. The NAP is not sufficient for a constitution. Ultimately, the best we can hope for is a consensus on principles among the residents of the land and all we can do now is attempt to establish a good, stable dynamic. As long as power remains distributed, things should work out.

That's why I'm beginning to think that lofty principles such as this might not even be helpful. If we require 3/4 of the Assembly to pass laws, allow the Assembly to rescind laws, have a sunset clause, and require each bill to be voluntarily paid in full, then we can hope that only laws with wide consensus will exist, and they will be funded by those who want them. At least we'll be trying something very new, that hasn't failed before.

If 3/4 of the people in Liberland want price controls, it won't be a very libertarian place, even with this bill of rights, just as the US isn't a very libertarian place despite it's bill of rights.

KacperZajc:
maybe we can qualify somehow what kind of rights we have in mind?

KacperZajc:
No law shall abridge the right of persons to the control of their own bodies and property, nor interfere with voluntary consensual or contractual relations among adult persons, nor impinge on the right to form cooperative ventures of any kind, so long as no NEGATIVE rights of others are infringed by force or fraud, or threat thereof, or unbeknown to them.

Does that deal with your fear of the right to free healthcare etc???

KacperZajc:
any comments?

KacperZajc:
Guys I would like to slowly wrap up our discussion, do you have any suggestions or comments of the last version I proposed?

KacperZajc:
does anyone object?

Jean-LouisMesic:
Hopefully, the duality of positive vs. negative rights will become more clear.

It is a very ambiguous, vague term for now.

PR #46, opened 2015-04-30
Constitutional prohibition of price controls

liberty90:
Direct prohibition may be useful. Probably nothing is more destructive for free market than price controls; prices are signals after all.

### DIFF
--- a/Liberland-constitution.md
+++ b/Liberland-constitution.md
@@ -34,6 +34,7 @@ The Bill of Rights shall constitute the integral part of the Constitution and sh
 * **§II.16.** No law shall oblige any person to acquire any form of insurance and/or pension scheme, nor shall it provide any incentive in that respect.
 * **§II.17.** No law shall prohibit video and/or audio recording of any employee of the Public Administration in public space and whilst on duty.
 * **§II.18.** No law shall abridge the right of self-defense against initiators of aggression, including the agents of the Public Administration where acting unlawfully and/or in error.
+* **II.19.** No law shall institute price controls of goods and services, nor rent controls.
 
 ### Article III: Individual Rights
 


KacperZajc:
I honestly love this idea.
I should have thought about this earlier :)

hazzik:
:+1:

Liberland-constitution.md
```diff
@@ -34,6 +34,7 @@ The Bill of Rights shall constitute the integral part of the Constitution and sh
 * **§II.16.** No law shall oblige any person to acquire any form of insurance and/or pension scheme, nor shall it provide any incentive in that respect.
 * **§II.17.** No law shall prohibit video and/or audio recording of any employee of the Public Administration in public space and whilst on duty.
 * **§II.18.** No law shall abridge the right of self-defense against initiators of aggression, including the agents of the Public Administration where acting unlawfully and/or in error.
+* **II.19.** No law shall institute price controls of goods and services, nor rent controls.
```
hazzik: "§" symbol missing

manchfu:
+1

joshmh:
I agree with the sentiment, but maybe we should consider a different approach. Can we distill lots of issues down to some fundamental rights? E.g, this addition should be covered under §II.3. Perhaps an addendum to the constitution which works through a number of examples of bad laws, such as this, and illustrates how they would violate §II.3 or other sections. I think this is a worthwhile exercise in any case. We should compile a list of all the bad laws we can think of, and see if they would be constitutional under this document.

KacperZajc:
I would rather make specific provisions.
General provisions will be too broad and therefore meaningless.
Id merge this one.

joshmh:
There's no way you can think of all the crazy things future politicians will want to do. I think we need to establish a few basic tenets, mostly the non-aggression principle, and then illustrate how everything we can think of fits into that.

KacperZajc:
The basic rule will be encapsulated in the provision where 'force and fraud' is proposed.
Here I want to list all we can think of to make sure it wont happen.

PR #47, opened 2015-05-01
New diagrams

burningtree:
I created new flowchart diagrams with [Gliffy.com](http://gliffy.com). Source and SVG version included.

@KacperZajc it's better? what do you think?


KacperZajc:
Awesome!
That is exactly what I had in mind :)

burningtree:
Good

PR #50, opened 2015-05-02
§II.13. Chemical substances and property rights

liberty90:
Maybe I do not understand current version right, but... I suppose that we do not want our homes damaged by polluted water from local chemical factory. I wouldn't be amused even in the case of purely unintentional contamination...


tinkyholloway:
I don't think your change covers the case of unintention damage to property as you intend, the clause at the end of the sentence takes the 'designed' context.

I think without this clause, which is intended I believe to cover recreational drug use mostly, that property owners would be protected by other areas of the C. By perhaps unambiguous wording of this clause, it does I think as you show, illustrate a problem.

I wonder if it is better to fix up the wording to cover recreational drug use more specifically, than it is to try and fix the problem by tying in the cases you highlight.

KacperZajc:
I think wording of this clause is good as it was intended to cover drug use only.

liberty90:
"intended". Obviously, but are you -sure- that this article couldn't be effectiviely used in court to argue about "rights" of polluting chemical factories ? xd

KacperZajc:
I think there is no harm in merging your proposal

Jean-LouisMesic:
Just in case I sail my ship using hemp rope and canvas, I don't want anyone smoking it, while I'm off-deck, amirite? : P

PR #54, opened 2015-05-07
Bills and their titles

liberty90:
This article was proposed previously on the liberland.org forum, not by me but by the user "kulibali". Seems quite sensible.


KacperZajc:
yea that is a good idea but I would include that in this provision:

Any Bill proposed to the Assembly shall be read aloud in entirety by the Assembly Speaker and shall contain rationale behind it and means of achieving thereof including precise calculations as to its potential impact on the budget; and shall specify under which power delegated by this Constitution to the Assembly it is proposed.

so:

Any Bill proposed to the Assembly shall pertain to one matter only as expressed in its title, it shall contain rationale behind it and means of achieving thereof including precise calculations as to its potential impact on the budget; and shall specify under which power delegated by this Constitution to the Assembly it is proposed; such Bill shall be read aloud in entirety by the Assembly Speaker before it is passed.

ok?

liberty90:
Sure, why not.

KacperZajc:
done

PR #56, opened 2015-05-09
No Jury Duty

jasoneriksen:
Jury duty is a form of slavery and has no place in a free society. Just like a military, juries can be staffed with voluntary jurors who can be enticed by any number of different reasons to volunteer for jury service.

### DIFF
--- a/Liberland-constitution.md
+++ b/Liberland-constitution.md
@@ -25,7 +25,7 @@ The Bill of Rights shall constitute the integral part of the Constitution and sh
 * **§II.7.** No law shall mandate any person to associate and/or transact with any other person; nor shall it prohibit it.
 * **§II.8.** No law shall abridge the freedom of thought and feeling, or their peaceful expression or dissemination, as in speech, press and other media, artistic depiction, or religious practice; nor shall any law promote or hinder any religion, artistic culture, gender or specific community.
 * **§II.9.** No law shall abridge the right to assemble peacefully where no rights of others are infringed.
-* **§II.10.** No law shall establish the institution of slavery, conscription, indenture, or any other form of involuntary servitude other than the Jury duty, within the Free Republic of Liberland, or in any place subject to its jurisdiction.
+* **§II.10.** No law shall establish the institution of slavery, conscription, indenture, or any other form of involuntary servitude, within the Free Republic of Liberland, or in any place subject to its jurisdiction.
 * **§II.11.** No law shall abridge the right of any person to use or issue any commodity or item as currency; nor shall the Public Administration of the Free Republic of Liberland engage in any form of monetary regulation or currency issue; nor shall it establish any Central Bank.
 * **§II.12.** No law shall propose, consent to, or request the incorporation of the Free Republic of Liberland, or any part thereof, to any other jurisdiction.
 * **§II.13.** No law shall abridge the use of any chemical substance by any person of age, so long as the use is not designed to harm any person without his or her consent and/or damages property without consent of the owner.


joshmh:
+1

KacperZajc:
Involuntary jury service might turned out to be necessary because:
1) you want criminal justice system, you must contribute (fair principle)
2) there might be not enough volunteers (risk of inability to go on trial)
3) volunteers are likely to be the same people over and over again (risk of manipulation)
4) volunteers are likely to be from lower classes (risk of bribery)
5) volunteers are likely to be from one social class (need for diversity)
6) volunteers might be signing up specifically to rule on some they know (risk of injustice)

I think there are enough reasons why we need this clause. Naturally, we will start with volunteers who we check and this is only in case we do not have enough people/diversity etc.

This is not a year long conscription, this is only several hours in total, with notice, and possibilities to be excused for valid reasons.

My impression is everyone here appreciates the importance of the jury in criminal trial so you should also appreciate the risks of leaving this entirely up to volunteers.

FellowTraveler:
> My impression is everyone here appreciates the importance of the jury in criminal trial

Yes, the importance of the Jury is to prevent the government from enforcing unjust laws. Otherwise the Judge is already qualified to render a verdict himself as we see in authoritarian countries.

jasoneriksen:
Each of the six points you outline could be used to argue for mandatory military service. The military is a necessary function of the government, just as a jury is a necessary function of the court system.

But neither organization requires slaves to function.

Rebuttals to each point:
1. I want a military to protect me from foreign aggression but that doesn't mean I should be compelled to serve. I want a criminal justice system because it is far better than the alternative, vigilante justice. How the government decides what resources are needed to implement the justice system is a separate question. It could do so through taxation as permitted by the constitution. It could do so by relying on donations. Both the military and criminal justice system are in place to protect my freedom, NOT to enslave me.

2. Then the trial will be delayed until there are enough volunteers (this happens even in countries where there is mandatory service). And there's an easy answer to this problem. Offer higher pay to jurors. Keep raising the wage and soon they'll be lined up out the door.

3. Risk of manipulation is not a legitimate reason to stop somebody from being a juror. If it can be demonstrated that a juror is corrupt then put them on trial and let the justice system punish them accordingly. If risk of manipulation were a legitimate reason, then no judge should be allowed to hear more than a small handful of cases before they must retire. No accountant should be allowed to handle the books for more than a few months before they must move onto another profession.

4. The principle is the same as point 3. Risk of bribery is also not legitimate. If you can demonstrate that a juror has been corrupted, put them on trial.

5. Jury diversity is not what's needed to have a well functioning jury that promotes justice. What is needed are objective jurors willing to listen to the facts of the case and decide on the facts. Rational people are needed. Intelligent people are needed. Honest people are needed. Independent thinkers are needed. Strong minded people of integrity are needed and these come from all walks of life. The justice system should only seek out jurors based on these requirements.

6. There's already a mechanism for dealing with this problem and it applies equally to a compulsory jury as it does to a volunteer jury. The judge asks the juror: Do you know the defendent? Do you have an ulterior motive? The judge and attorneys can decide whether or not they think the juror should sit on the panel.

I am not challenging the value of a trial by jury. I think that is a fine thing to include in the Constitution.

I think most people agree that Liberland should have a police force. Should mandatory service in the police be a requirement (I would pose your same 6 points to this question)?

FellowTraveler:
@jasoneriksen Why not just say that registering to vote, and registering to be available for jury duty, are one and the same thing? That way anyone who has registered to vote has automatically volunteered for jury duty.

KacperZajc:
@FellowTraveler  I think this is the best idea here.

jasoneriksen:
@FellowTraveler I agree with you that protecting against unjust laws is one important function of the jury, especially in countries with unjust laws (that includes just about every country, except hopefully Liberland) but I don't think it's the only important function of the jury.

I do view your solution of tying voting rights to jury duty as MUCH better than mandatory jury service, but I still would disagree with it as the best solution. By that argument the right to vote could be limited only to those willing to serve their country in other ways, such as through military service. The right to vote in elections which determine who will be trusted to run the government (even a strictly limited government) is an important right for a citizen and it should not come with any strings attached.

jasoneriksen:
I would further pose this question. Which scenario provides the most freedom in a republic where the government is determined by an election where citizens vote:
1. Mandatory jury service of all residents
2. Mandatory jury service of all Voting Citizens
3. Voluntary service.

Without getting into details about how this would be implemented, I can look at those 3 options and say, number three promotes freedom more than the other two. Implementation is how that jury system should be managed is up to whoever ends up running the government, and is beyond the scope of this Constitution.

KacperZajc:
you know what promotes more freedom? no state whatsoever
level of freedom is the most important criterion but not the only one
this state must work, this is not anarchy
I think we should link jury service to voting rights
this gives more liberty, yet is fair

jasoneriksen:
Anarchy does not promote more freedom. If you define maximum freedom as the ability to live your life with as little coersion from other people as possible, then anarchy actually is quite a problem. Under anarchy, gangs, warlords, and violence are the norm. These are the antithesis to freedom.

The state must work, yes. But the job of the state is to protect the rights of the people. Jury Duty undermines those rights. Jury duty tied to voting is a better alternative than compulsory service, but don't call it the right to vote anymore. Now it's a privilege granted by the state that comes to citizens who volunteer to serve on juries.

Is it fair that a person should be allowed to vote on who will run the government responsible for protecting them from criminals and invaders? I think that's fair.

Should that ability to vote be tied to some mandatory term of service? That is not fair.

There are plenty of ways to make the state work without making jury service mandatory. I can think of 3 right off the top of my head:
1. Pay the jurors a slightly above average wage than what they would get working in Liberland.
2. Promote jury work at colleges as an awesome way to see first hand how their government treats those suspected of crimes, not to mention an opportunity to make your opinion heard. You think your vote counts in an election where tens of thousands are voting? How about in an election where 12 jurors are voting on the fate of a single person? 
3. Offer to compensate employers for the time off their employees will need. Most companies in Liberland will be run by proud citizens of Liberland, who would view it as a mark of pride to be engaged in the civic process.

I'm not a human resources specialist or a recruiter, but it's obvious there are ways to get an adequate supply of qualified jurors to participate in the process voluntarily. 

And finally. Who makes a better juror? The person who was forced to serve against their will, or coerced to serve as a condition of being allowed to vote? Or the person who volunteered to be there?

Volunteer jurors will be more motivated to seek justice.

FellowTraveler:
I have to admit your arguments are compelling. And if it works -- which it appears that it would -- it would really clarify the slavery inherent in jury conscription in other countries.

joshmh:
Also, jury duty in the US sucks for the juror. It's a huge waste of time because there's a good chance you won't even make it to the trial, there's a lot of waiting around and bureaucracy, and the whole thing is usually pretty boring. If you make the experience a lot more fun, it would make it easier to get people voluntarily.

KacperZajc:
I can remove that part form this article but this will not preclude linking the jury duty with voting rights later

PR #60, opened 2015-05-09
fair representation

FellowTraveler:
The U.S. Constitution provides for 1 representative for every 30,000 people. Unfortunately, Congress froze the size of the House of Representatives in the early 20th century. As a result, the people's vote for their representative has been steadily diluted over the past 100 years. Today there is 1 representative in the USA for every 700,000 people.

Let's be clear in understanding that 1 representative for 10,000 people is fair representation, whereas 1 representative for a million people is tyranny. To protect against this "tyranny from dilution" we must insure a minimum number of representatives for every X number of citizens.


Liberland-constitution.md
```diff
@@ -71,13 +71,13 @@ The Free Republic of Liberalnd shall be governed by the Public Authority in acco
...
 * **§IV.7.** The power to initiate a legislative procedure shall lie with the Cabinet only.
-* **§IV.8.** An ordinary Bill shall be passed into law with the majority of no fewer than fifteen votes of the Assembly Representatives subject to the power of general veto by the Citizens as provided in §IV.13 and §IV.14.
+* **§IV.8.** An ordinary Bill shall be passed into law with the majority of no fewer than three-quarters of the votes of the Assembly Representatives subject to the power of general veto by the Citizens as provided in §IV.13 and §IV.14.
 * **§IV.9.** A Constitutional Bill proposing an Amendment to this Constitution shall be passed only with the unanimous vote with all Assembly Representatives present and participating in the vote, and shall be subject to the power of general veto by the Citizens as provided in §IV.15.
 * **§IV.10.** The Financial Bill containing the provisions of the Annual Budget shall be passed with the majority of no fewer than eleven votes and shall be subject to the power of general veto by the Citizens as provided in §IV.15.
```
jasoneriksen: If the number of Assembly Representatives is fluid, this number needs to be too.

FellowTraveler: Which number? I already changed "fifteen" to "three-quarters".

Unless I'm missing something...?

Liberland-constitution.md
```diff
@@ -71,13 +71,13 @@ The Free Republic of Liberalnd shall be governed by the Public Authority in acco
...
 * **§IV.6.** If no Chancellor is elected in accordance with the procedure provided in §IV.4 for more than ninety days, the General Election must be held.
 * **§IV.7.** The power to initiate a legislative procedure shall lie with the Cabinet only.
-* **§IV.8.** An ordinary Bill shall be passed into law with the majority of no fewer than fifteen votes of the Assembly Representatives subject to the power of general veto by the Citizens as provided in §IV.13 and §IV.14.
+* **§IV.8.** An ordinary Bill shall be passed into law with the majority of no fewer than three-quarters of the votes of the Assembly Representatives subject to the power of general veto by the Citizens as provided in §IV.13 and §IV.14.
 * **§IV.9.** A Constitutional Bill proposing an Amendment to this Constitution shall be passed only with the unanimous vote with all Assembly Representatives present and participating in the vote, and shall be subject to the power of general veto by the Citizens as provided in §IV.15.
```
jasoneriksen: I'm not saying if this is a good or bad thing, in fact it might be an awesome positive effect of this change, but it would become next to impossible to achieve a unanimous vote on just about anything once the Assembly grows to a few hundred members. This in turn would make a constitutional amendment next to impossible.

This could be very good, because it would make the legal erosion of the freedoms spelled out in the Liberland Constitution next to impossible.

But it would also make any amendment to the Constitution effectively impossible.

Liberland-constitution.md
jasoneriksen: Line 82 says a majority of no fewer than eleven votes. What does that mean after the Assembly has grown to 100 members? If only 20 of them show up, and 11 vote for the Financial Bill, is that enough to pass it?

jasoneriksen: something like: "with the majority of no fewer than one-half of the votes of the Assembly Representatives"

Liberland-constitution.md
FellowTraveler: Thank you for the "eleven" thing. Fixed.

As for your other questions, they are valid but should be addressed in a separate pull request or issue.

I would change the "unanimous" requirement to a super-majority, and I would also add that any Constitutional amendment votes must be separated in time. For example, a super-majority must approve it once, and then again a year or two later, before it's ratified. Similarly, a supermajority of the Citizens must vote to approve it, and then must vote AGAIN to approve it a year or two (or five) later, before it's actually ratified.

"Present and participating" probably refers to a requirement for a quorum. But again, these can all be addressed in a separate PR or issue.

jasoneriksen:
:+1:

Liberland-constitution.md
jasoneriksen: Agreed that these matters are best dealt with in a separate PR if needed.

FellowTraveler:
The original text read that there are "20" representatives, and that "11" or "15" must vote to do certain things.

I changed "11" and "15" to "a majority" and "three-quarters". Unfortunately, I did not specify that it was "a majority of ALL Assembly members" and "three-quarters of ALL Assembly members."

This created a danger that a "majority" could occur with less than 11, if some of the Assembly members failed to participate in the vote.

Therefore I have fixed this problem with this PR, in order to retain the original intent. Now it requires "a majority of all Assembly Representatives" and "three-quarters of all Assembly Representatives" which is effectively 11 and 15, and will work correctly even when there are more than 20 Assembly members.

KacperZajc:
at this point we dont know what kind of representation we want. 20 was the number President himself requested. It might grow up in the future but for now it stays what it is. Im closing this pull.

joshmh:
@KacperZajc, I think we should be encouraging people to engage in thoughtful debate. Unilaterally closing discussions without allowing for serious discussion could drive off the best contributors and undermine the legitimacy of this document. Personally, it doesn't really interest me what the President has requested. I'm only interested in exploring these intellectual pursuits.

FellowTraveler:
This is the sort of issue that will have a huge impact on the country's future. If you have 20 Representatives with 200,000 Citizens, then that is fair representation. But if you have 20 Representatives with 20 million Citizens, then it is tyranny (1 ruler over 1 million people.)

Therefore the way the Constitution is currently structured, it will start out with fair representation, but then the power of the people will slowly be diluted over time until their voice in the Assembly has been lost.

This is what happened in the USA. There used to be 1 representative for every 30,000 people. If that were true today, he could not get elected unless he lived in your town and operated a business down the street. No amount of TV commercials could change that. He would have to meet with you, and listen to your concerns, or you would fire him.

But now, since Congress froze the size of the House (just as the Liberland Constitution freezes the size of the Assembly) there are now 700,000 constituents per representative. Most of them have never met their representative. So then who does meet with him? Large corporations and public employees unions. They pay for his TV commercials that get him elected, and they dictate the contents of the legislation that gets passed.

The people in America have lost their representation and the same will eventually happen in Liberland unless you get this one right.

joshmh:
Another example. In the State of New Hampshire, there is one representative for every 3,500 people. The representatives are your friends and neighbors and they don't get paid or even have offices. It works rather well.

mtelka:
Is there any justification/need to create the Assembly at all?  Assuming there is just a rare need for voting (either new/changed acts, or changes to the Constitution, etc.), wouldn't be better (more stable/fair) to let to vote all Citizens?  The omission of the Assembly would also simplify the Constitution, limit the "government", prevent the tyrrany (mentioned above), etc.

FellowTraveler:
> Is there any justification/need to create the Assembly at all?

An assembly is a deliberative body, and it's more likely that all voters in the assembly are educated on the issue and have debated it. They also have their reputations tied up in their voting record.

Often in direct democracy we see a "tyranny of the majority." I personally prefer a system where the citizens have an automatic veto (as seen in the Liberland Constitution) so they have the opportunity to prevent bad laws from being imposed, but without having the power to impose bad laws themselves.

mtelka:
> An assembly is a deliberative body

Please help me to understand what do you mean by "deliberative" in a light of IV.1.:

> "the Assembly", which shall exercise it on behalf of the Citizens.

Apparently, the proposed Assembly have the legislation power (I mean: a power to vote for new law/act).  This does not sound as a "deliberative body".

KacperZajc:
@FellowTraveler I agree and that is why they have the power to veto yet they do not have the power to initiate the legislative procedure

mtelka:
> it's more likely that all Representatives in the assembly are educated on the issue and have debated it

How this differs to all parliaments around the globe?  With this reasoning you do not need Liberland.  Your current country (likely) follows that too.  Are you happy with your parliament in your country?

mtelka:
> Often in direct democracy we see a "tyranny of the majority."

How the indirect democracy prevents the "tyranny of the majority"?

Please note that I'm not a fan of the direct democracy.

KacperZajc:
@mtelka if you suggest we dont need to make any laws, this is ridiculous; if you suggest another body can make laws instead of assembly this is against the separation of powers which one of the most important principles regarding governance., so Im not sure what are you trying to do here.

mtelka:
@KacperZajc no. I do not suggest that.  I suggest to leave all the voting power to Citizens.  I would leave the power to initiate a legislative procedure to Cabinet (IV.7.).  My suggestion is based on the fact that I do not see any pros to have the Assembly.  The absence of the Assembly will make the law changes slower, and harder (and hopefully less bribery prone), so the whole law of the countly would be more stable and predictable.

For example, if the Constitution would say that to accept a new law you need more than a half of the Citizens voting for it, it would be very hard to accept a law wanted by just 11 people in the Assembly.

KacperZajc:
I think people should take care of their business and live their lives not spend hours analysing legislative proposals.
Plus, as I have been saying repeatedly here, we need to look like a proper state.

FellowTraveler:
@mtelka "Your current country (likely) follows that too. Are you happy with your parliament in your country?"

Well I am in Switzerland and the legislative bodies here are not so bad. They have fair representation, and there is competition between the cantons.

Unfortunately I am from America and I am extremely unhappy with the Congress there. (In America there is no more fair representation; the people's votes have been diluted to 700,000 to 1.)

In fact, **this is why I have submitted this pull request:** because Liberland has no fair representation. The Assembly being frozen at 20 Representatives will eventually result in diluted representation just as we see in America.

Furthermore, Liberland has no competition between cantons either. (I submitted an issue about this as well.)

mtelka:
> I think people should take care of their business and live their lives not spend hours analysing legislative proposals.

That's exactly my point.  With the direct voting nobody would vote for often and complex law changes, so the Cabinet will need to propose simple, clear, and infrequent law changes.  I believe we all have some experience with the fast and complex law in our current countries...

> Plus, as I have been saying repeatedly here, we need to look like a proper state.

Is there any definition of the "proper state"?  Is some form of parliament needed for the "proper state"?

KacperZajc:
@FellowTraveler I assume we will get back to the fair representation topic once Liberland's population grows.

@mtelka proper state here in a sense that it has 3 powers which balance and check each other.

mtelka:
@FellowTraveler I understand why you suggested this changeset (I read all the comments above).  I just want to discuss even more fair representation (not 700,000 to 1, not 10,000 to 1, just 1 to 1).

@FellowTraveler How the Assembly containing all the Citizens violates this "3 powers rule"?

FellowTraveler:
@KacperZajc:
> I assume we will get back to the fair representation topic once Liberland's population grows.

I think it will probably be too late. Right now the Constitution is being formed, it is the best time to get it right up front.

People in Liberland currently do not have a right to fair representation.

In America they froze the size of the House, and it has remained frozen for a hundred years!

But you think you will fix it "later" when it will require a Constitutional amendment to do so?

Once there is 1 Representative for every million people, they will be so drunk with power, they will never let things change. They will be able to do whatever they want with impunity, and the large corporations and unions will control everything, just like in America.

These are the subtleties that are used to destroy liberty.

KacperZajc:
This is not my call, I was instructed to put there 20 by the President.

FellowTraveler:
@KacperZajc
> This is not my call, I was instructed to put there 20 by the President.

Please make sure he sees this thread.

BTW I think the Assembly members should be able to propose legislation, not just the Cabinet. The Cabinet will be focused on executive duties. The purpose of the Assembly should be to propose and debate legislation, and that is safe as long as we have the Citizens' veto and the Bill of Rights.

KacperZajc:
the fact that only Cabinet can propose legislation was well received by everyone here; it is designed to limit the volume of legislation (only cabinet only once a year) and while the cabinet can exist with 11 votes, bill doest pass with fewer than 15 so there is a strong separation of power - the cabinet will not control the leguslature

mtelka:
@KacperZajc only once a year?  I'm unable to find it in the Constitution.

KacperZajc:
§V.3. The Chancellor, on behalf of the Cabinet, shall once a year publicly give to the Assembly information on the State of the Republic, and recommend to their consideration such measures as the Cabinet shall judge necessary thereby initiating a legislative procedure in accordance with the provisions of Article I.

mtelka:
@KacperZajc IMHO, that does not prevent the Cabinet to propose a new Bill more often than once a year (let say, every week).

KacperZajc:
we can clarify if there are any doubts

FellowTraveler:
Bills will still originate in the Assembly regardless. The reason is because 3/4s of the Assembly is needed to pass any bill. This gives them the power to make deals with the Cabinet. **Everything will get back-channeled.** Representatives will just vote against the administration's bills unless the administration also proposes the Representative's bills so the Assembly can vote on them. I assume the Assembly also has the power of the budget, which they will also exercise for this purpose.

Also -- when a Citizen proposes necessary legislation, he would normally contact his Representative to discuss it. (Assuming he has fair representation -- otherwise he has no one to contact!) But from there the Representative can do nothing since he cannot propose legislation.

The Citizen cannot realistically be expected to contact the President about such matters. Even now the President probably won't even read this thread, even while the population is so small that the country's still mainly populated by boars and birds.

Perhaps Citizens *should* be able to propose legislation. If they can gather enough signatures, they should be able to put it to the Assembly, and then if it passes, it goes to Citizens' veto.

mtelka:
@KacperZajc If the intent is to propose new Bills just once a year, then the better wording is definitely needed.  OTOH, why is it better to see proposed bunch (let say dozen) of new Bills once a year, instead of one new Bill a month?  Or, is the intent to propose just exactly one Bill per year?

mtelka:
@FellowTraveler I agree with you, but I suggest to talk about the voting only here (and the size of the Assembly), not the legislation proposals.  Yes, my previous comment is off-topic too, so I apologize.

KacperZajc:
The size of the assembly is out of question for now unless President decides otherwise, I'll ask.
The cabinet stays as the only source of legislative proposals because it reduces the volume of laws and everyone was happy with this idea - we will only clarify that it is possible only once a year so that there are no doubts.

mtelka:
BTW, the size of the Assembly written to the stone as 20 is not robust for several reasons (corner cases, but still):

1) In a case there is less than 20 Citizens you won't be able to get 20 Representatives.
2) In a case there is slightly more than 20 Citizens and (almost) all of the Citizens apply for the candidacy to Assembly, you could easily get, let say, 30 candidates with one vote.  Again, you won't be able to establish the Assembly with 20 Representatives.
3) Generalizing the 2) above.  In the General Election you could get more than 20 winners of the election with the same number of votes.  This should be resolved somehow (probably not in the Constitution, but in some election bill).
4) What in a case there is less than 20 candidates in the General Election?  Again, you won't be able to establish the Assembly with 20 Representatives.

PR #61, opened 2015-05-09
Assembly members cannot exempt themselves.

FellowTraveler:
In the USA there is a huge problem with members of Congress exempting themselves from the laws that they pass.

There has also historically been a number of problems caused by the selective enforcement of laws.

The purpose of this pull request is to provide equal protection under the law, and to prevent Assembly members from exempting themselves from laws.


KacperZajc:
I like your idea here but I would only mention exemption by the Assembly Representatives because 'group of interests' is too wide

FellowTraveler:
Okay I added the fix you requested.

KacperZajc:
no, I dont know why it shows it here.
I removed the drug protection today because the President said it was unsustainable because of PR (see a separate issue on that).

make it Assembly Representatives, not Assembly members

FellowTraveler:
I will close this pull request and add the fix you requested in a new pull request.

FellowTraveler:
Discussion continued here: https://github.com/liberland/constitution/pull/70

PR #62, opened 2015-05-09
removed unnecessary wording

FellowTraveler:
The wording seems superfluous. By removing it, the paragraph becomes tighter and harder-hitting.

Removed: "attempting a harmonious consensus for the benefit of all those seeking their home in this place"


jasoneriksen:
:+1:

KacperZajc:
this wording is a fruit of a long discussion everyone here took part in, it stays for now

PR #70, opened 2015-05-09
No exempting Representatives.

FellowTraveler:
No exempting Representatives from laws.


FellowTraveler:
This is a new pull request that continues from https://github.com/liberland/constitution/pull/61

PR #77, opened 2015-05-10
eleven/fifteen

FellowTraveler:
I changed *"vote of eleven"* to be *"vote of a majority of all Assembly Representatives."*

I also changed *"vote of fifteen"* to be *"vote of three-quarters of all Assembly Representatives."*

NOTE: this PR does *not* change the operation of the Constitution at all. **The behavior is identical.** The purpose of this PR is to make it easier later to make changes in §IV.2 with a future pull request.

This PR stands on its own, and since it doesn't change the functionality of the Constitution in any way, it should be fine to merge it. This way any future debate can occur when a future PR is submitted for §IV.2, and it won't require changes all over the Constitution, but will localize any changes to that specific provision.

### DIFF
--- a/Liberland-constitution.md
+++ b/Liberland-constitution.md
@@ -73,20 +73,20 @@ The Free Republic of Liberland shall be governed by the Public Authority in acco
 * **§IV.1.** All legislative powers of the Free Republic of Liberland shall be vested in the Citizens of the Free Republic of Liberland and the Assembly of the Free Republic of Liberland, hereinafter, "the Assembly", which shall exercise it on behalf of the Citizens.
 * **§IV.2.** The Assembly shall consist of twenty Assembly Representatives elected by the Citizens in the General Election held every four years.
 * **§IV.3.** No Assembly Representative shall hold the office for more than eight years.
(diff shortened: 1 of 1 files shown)


KacperZajc:
I understand the purpose, I dont mind :)

KacperZajc:
wait that poses a problem:
this either make it necessary for all representatives to attend the meeting OR allows to pass with a majority of representatives present, no matter how many of them it is
am I right?

FellowTraveler:
The idea is to make it behave the same as before, without having to state the number.

For example, if there are 20 representatives, then the "majority of ALL representatives" would be 11. This doesn't mean that all 20 actually have to attend the meeting, but it does mean that at least 11 of them must vote for the provision.

Someday in the future, let's say there are 50 representatives. In that case, the "majority of ALL representatives" would be 26. This doesn't mean that all 50 actually have to attend the meeting, but it does mean that at least 26 of them must vote for the provision.

KacperZajc:
aaah ok this interpretation makes sense

cnavigato:
interesting change, well thought out ;)

PR #78, opened 2015-05-10
Fair Representation

FellowTraveler:
This pull request will result in 20 Assembly Representatives until there are over 200,000 Citizens.

After that, there will be 21 Representatives for 210,000 Citizens, 22 Representatives for 220,000 Citizens... 50 Representatives for 500,000 citizens, and so on.

Without this change, the vote of the people will be steadily diluted over time as the number of Citizens grows. **We have seen this problem in the United States, where each Representative "represents" over 700,000 people, and the people thus no longer have fair representation in Congress.**

Once the vote of the people becomes diluted, it may become impossible to fix this later. Better to make sure they never get diluted in the first place.

### DIFF
--- a/Liberland-constitution.md
+++ b/Liberland-constitution.md
@@ -88,7 +88,7 @@ The Free Republic of Liberland shall be governed by the Public Administration in
   * **§II.1(2)** The Assembly may extend or shorten its session with a simple majority vote of the overall number of Assembly Representatives.
   * **§II.1(3)** The Assembly may reconvene at any time whatsoever upon the call of the Assembly Speaker.
   * **§II.1(4)** The Assembly shall not be dissolved, nor shall its session be interrupted, by any other branch of the Public Administration.
-* **§II.2.** The Assembly shall consist of twenty Assembly Representatives.
+* **§II.2.** The Assembly shall consist of at least twenty Assembly Representatives. The Citizens having a right to fair representation, there shall be at least one Assembly Representative for every ten thousand Citizens.
   * **§II.2(1)** All Assembly Representatives shall receive symbolic remuneration for their service which does not constitute a significant burden on the Annual Budget and does not exceed 5% of the Annual Budget spending in any event.
   * **§II.2(2)** No Person shall hold the office of an Assembly Representative whilst holding any other public office within the executive or judicial branch of the Public Administration concurrently.
   * **§II.2(3)** No Assembly Representative shall be precluded from taking part in any Assembly vote and/or debate by virtue of being detained prior to his or her trial.


KacperZajc:
as I said I could not authorise such a change because I had been directly instructed by the President to set the number at 20; maybe he plans on increasing it over time, I'll ask

FellowTraveler:
The number is definitely 20. It will only go up in the future after there are hundreds of thousands of Citizens, and only as necessary to preserve their right to fair representation. (Otherwise you couldn't really call them "representatives" anymore -- they certainly don't represent anyone in the USA anymore except large corporations and public employees unions.)

I think the President will be fine with this change if you show him this PR. I hope, anyway.

joshmh:
:+1: 

The only way the president could object to this is if he wants to keep 20 assembly reps even when there are over 200,000 citizens, which means Liberland will be covered with skyscrapers.

FellowTraveler:
We know for a fact that it becomes tyranny at *some* point. For example, 20 Representatives over 20 million people, would be 1 representative per 1 million people.

...But that is not a representative at all, but a king!

Zephram-C:
Just to give some perspective. If the population density in Liberland is ... then its capacity is ... (approximately).

Density like | Liberland capacity
------------ | -------------
USA | 200
World | 325
Japan | 2000
Taiwan | 3900
Bangladesh | 6400
Vatican City | 11 000
Hong-Kong | 38 000
Moscow | 63 500
Monaco | 110 000
Paris | 129 000
Manhattan | 156 000
Ho Chi Minh City - District 10 | 244 000

*But there is a possibility of on-line citizenship without actually residing in the place.*

FellowTraveler:
Yes there is the possibility of online citizenship -- and there is also the possibility that if the country is successful, that it will grow in size.

joshmh:
@Zephram-C Nice table! I agree with @FellowTraveler, if it reaches 100k people with skyscrapers, it's very likely it could expand.

jasoneriksen:
:+1: And I would be open to reducing the number even further to one representative per 1,000 citizens. Liberland doesn't need to be encumbered by brick and mortar assembly meetings that tend to limit the practical size of a legislature (though a building with tens of thousands of assembly representatives debating and arguing would be impressive). The Assembly can meet virtually, even debating legislation on a forum such as Github in a public and accessible format.

FellowTraveler:
https://www.youtube.com/watch?v=Lo4cFViNLes

mtelka:
@jasoneriksen :+1: For example, in Czech Republic the number is about one representative per 52,000 citizens (considering the lower house of the parliament only), or even 37,000 (considering all the parliament) and it is bad, bad, very bad (read: definitely not enough).  Similar situation is in Slovakia too (one representative per 36,000 citizens) and it is not enough as well.

So the suggestion for the Liberland Constitution to have one representative per 10,000 citizens could sooner or later lead to very unsatisfying situation.

I support to change the number to 1,000 citizens (or even lower number).

jasoneriksen:
@FellowTraveler :+1: Can we put in a clause that the Chancellor must wear robes like Palpatine?

KacperZajc:
@jasoneriksen to be honest I was thinking about the 18th century clothing for all Assembly Representatives but that's just me... :)

jasoneriksen:
@mtelka This is interesting. Thank you for sharing that perspective. 

Is there any reason to object to a direct 1:1 democracy (with strong constitutional protections)?

@KacperZajc Your idea is better. 18th century fashion was far superior to the look of the Star Wars prequels.

mtelka:
To make it clear:  I support this PR.  We could consider this Fair Representation change in the Constitution as the first step.  We could improve the number later.

mtelka:
@jasoneriksen I suggested that in PR #60, but it was denied :-(.

yamamushi:
:+1:

Jean-LouisMesic:
+1

I second the motion to cap the constituent to representative ratio at 10k.

FellowTraveler:
There is further discussion on this in #60

PR #80, opened 2015-05-11
sense of justice

FellowTraveler:
Changes *"right to acquit"* to *"right to acquit based on personal sense of justice"*.

The purpose of this PR is to restrict the Jury's acquittal powers only to cases where they believe injustice would otherwise be served. Otherwise they should rule according to the law as advised by the judge.

I prefer *"based on conscience"* and I believe there is a strong legal tradition behind the word *"conscience"*. But I noticed that *"sense of justice"* is already used in the laws of Liberland: https://github.com/liberland/laws/blob/master/drafts/Adjudication_of_Civil_Disputes.md

History shows that juries use this power very rarely, but when they do, they do so correctly, and it protects against tyranny.

@KacperZajc your main concern was that jury prejudices might be excited. For example, you said that (not to offend anyone) a muslim juror might acquit a jihadi. But I believe this will not be a problem, for three reasons:

1. Muslims convict jihadis of crimes all the time.

2. (More importantly) The Constitution already guarantees an **"impartial jury."** This means that in your example, a muslim with strong feelings about jihad would already have been excused during the jury selection process.

3. The *entire jury* would have to choose to acquit. That's twelve dissenters based on principle, not just one. And chances are, if the entire jury believes the government is perpetrating an injustice, then the jury is probably right.

The right of the Jury to acquit when they believe conviction will be an injustice, has a long history in the English legal tradition, and in fact is the reason why we have juries at all. I hope that you will consider the importance of this change to the future of liberty in Liberland.


yamamushi:
+1

Jean-LouisMesic:
+1

I second this.

FellowTraveler:
The new PR is #344

PR #82, opened 2015-05-13
The Constitution should not depend on technologies

mtelka:
I believe the Constitution should not depend on technologies.

**Rationale:**
The communication technology is changing and developing very fast.  We cannot expect that the current level of the technology won't ever cahnge.

Imagine Liberland is formed in 1800 and we want to allow all Liberland Citizens around the world to participate on referendum.  If we would want to specify in the Constitution the best communication technology we could find in 1800 we would very likely use something like this:

* **§IV.16.** Any referendum in the Free Republic of Liberland shall be conducted via the postal service.

After about 50 years (in 1850) we would realize that nobody wants to use slow and unreliable postal service when we do have the fast and cool new technology: telegraph.  So we would want to amend the constitution to read:

* **§IV.16.** Any referendum in the Free Republic of Liberland shall be conducted via the telegraph.

But after some years (let say in 1920) we would again touch the Constitution to read:

* **§IV.16.** Any referendum in the Free Republic of Liberland shall be conducted via the Bell's telephone.

In 1970 we would amend the constitution again to read:

* **§IV.16.** Any referendum in the Free Republic of Liberland shall be conducted via the facsimile or telephone.

And we are satisfied now (in 1970), because we are able to deliver very fast the exact copy of the ballot from anywhere directly to the Liberland election commision.  Wow!  Incredible!

But, wait, in 2005 we would realize that nobody uses fax and telephone is... so legacy.  And, we do have the Internet around!  So, amend the Constitution to read:

* **§IV.16.** Any referendum in the Free Republic of Liberland shall be conducted online via the Internet.

What will happen in the future?  We will need to amend the Constitution again once the teleport is finally invented, then again, once the telepathy starts to work reliably, and again, again and again.  Just to have the latest and greatest technology listed in the Constitution as a communication medium for the referendum.  Why?

I believe the Constitution does not need to specify such technical details as we see in **§IV.16.**.  The technical details about the referendum should be in a separate referendum Bill.

Note: We should also allow General Election over Internet, but this does not belong to the Constitution as well.


KacperZajc:
I agree, I put it there because the President mentioned it but you are right it can be implied only.

mtelka:
If there is no objection, please merge the PR.  Thanks.

KacperZajc:
you have to renumber

mtelka:
Hm. I followed CONTRIBUTING.md...
But not problem, I'll renumber.

mtelka:
Renumbered...

KacperZajc:
ok, does anyone object? 
(give a couple of hours)

liberty90:
No objections, on the contrary - this is very good pull request. Technologies change.

burningtree:
:+1:

PR #85, opened 2015-05-15
After a line is removed you need to renumber

mtelka:
Based on my experience with PR #82, the contributing rules for line removal needs to be updated.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,7 +20,7 @@ If you want to add a new, numbered line to the constitution, add it as the last 
 
 ## Removing a line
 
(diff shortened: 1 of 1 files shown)


hazzik:
I disagree. When any bill (and constitution) will became in power paragraphs shall never be renumbered. Otherwise you will need to update all the official documents (including court decisions on offences) which reference the amended document to reflect these changes.

KacperZajc:
I assume he means now before it is published in a final version?

mtelka:
@KacperZajc yes, exactly.  The CONTRIBUTING.md file is describing the current process of the community effort to improve the Constitution draft.

I agree with @hazzik that once the Constitution is in force we should not renumber everything in a case we will remove something.  But that is not the current situation now.

Issue #93, opened 2015-06-06
Prosecutor 


saintego:
How is Chef prosecutor appointed? Will he be able to refuse to prosecute somebody? Because this way it is easy way to avoid court, if you are friend of prosecutor.

saintego:
I made pull request #94, but maybe shorter period will be better

PR #94, opened 2015-06-06
Appointing of The Chef Public Prosecutor #93

saintego:
Maybe life period is too long and 4,5 or 6 years would be better


saintego:
I wanted to open discussion, because I think Constitution should have that point in the way it is written

KacperZajc:
I agree it needs discussion, I want to hear your opinions here.

liberlandcitizen:
Term limits seem reasonable for all public offices with power.

saintego:
I think it's general practice that prosecution have different period with government(so 5 or 7 years)  but I think everything depends on power of prosecutor. Will he be able to cover crime by not to process it to court?

Issue #101, opened 2015-06-09
List of potential changes for the Draft 3.0 IMPORTANT


KacperZajc:
1.	Add to the Bill of Right’s preamble that it applies also regardless of sexual orientation.
2.	Make the Justices of the Supreme Court who are to replace the ones removed in a referendum directly elected.
3.	Oblige the Cabinet to consider a bill proposal put forward by 5 Assembly Representatives (no obligation to actually submit such a bill) 
4.	Remove the word ‘indenture’ from II.4 because as someone pointed out there were forms of voluntary indenture and the involuntary ones are already covered by the article.
5.	Add a possibility to issue a warrant where there is a risk the accused will flee or interfere with the investigation (not only pose a threat to others) in III.4.
6.	Add a ban on being a judge and an Assembly Representative at the same time.
7.	Qualify the term for the Chief Public Prosecutor (see #94)
8.	Add a ban on requirement to identify oneself (see #64 ).
9.	 Add a ‘persons of age’ qualification to II.3 (ban on victimless crimes).
10.	Clarify that agents of the Public Administration can be required to be adequately educated and qualified (in II.12).
11.	Clarify that ban on regulation of marriage does not preclude the Assembly from regulating inheritance rules where there is no will.
12.	Remove word ‘associate’ from II.6 because it might preclude the court from issuing injunction for stalkers etc. plus the word ‘transact’ is enough.
13.	Clarify that the right to issue anything as commodity does not apply to ‘things’ which are illegal (e.g.: add "unless either party of transaction is legally prohibited to possess such item or commodity" to II.10).
14.	Add protection ‘of others’ to the right of self-defence and defence (in addition to the right to defend ‘the close ones’ in II.11).
15.   Consider whether 1 year limitation for prosecution is appropriate (see #99 )
16.   Lift the threshold for constitutional amendment (and taxation if remains) to 3/4 in a (mandatory) referendum.
17.	Restate the right to bear arms (as partially proposed by the gun expert, criminologist and a person who will be responsible for the future gun laws): 

"No person of age shall be deprived of the right to own, manufacture, sell, transport, and bear any small arms unless decided otherwise by a decision of the court where law allows it as punishment for crime, or on medical grounds; licensing shall not be used to infringe upon this right, nor restrictions as to the purchase and/or use of ammunition and/or the firearms accessories.” 

[...]

PR #104, opened 2015-06-09
Expanded right to bear arms

terrorist96:
Made it so arms accessories and ammunition are also protected rights.


ke7ofi:
What if we were to make sure the right to self-defense and necessary instruments is the thing protected and not the physical objects? That would simplify things a lot.

terrorist96:
I propose we use the Missouri constitution as guidance on this. Please see the full text of the arms amendment:

That the right of every citizen to keep and bear arms, ammunition, and accessories typical to the normal function of such arms, in defense of his home, person, family and property, or when lawfully summoned in aid of the civil power, shall not be questioned. The rights guaranteed by this section shall be unalienable. Any restriction on these rights shall be subject to strict scrutiny and the state of Missouri shall be obligated to uphold these rights and shall under no circumstances decline to protect against their infringement. Nothing in this section shall be construed to prevent the general assembly from enacting general laws which limit the rights of convicted violent felons or those adjudicated by a court to be a danger to self or others as result of a mental disorder or mental infirmity.

KacperZajc:
see #101 it is included

PR #122, opened 2015-06-13
Pursuit vs pursuance

terrorist96:


terrorist96:
Google says that "in pursuance of a warrant" is a more common phrase than "in pursuit of a warrant".
The former means you're acting in accordance to a warrant, the latter means you are attempting to procure a warrant. I think that all instances of this phrase should be changed to the former. Thoughts?

KacperZajc:
or pursuant to ?

terrorist96:
"Pursuant to" and "in pursuance of" have the same meaning, just different phrasing. I'm ok with either.

KacperZajc:
I will change it everywhere to "in pursuance of" :)

KacperZajc:
I will change it everywhere to "pursuant to" then :)

terrorist96:
Cool. :smile:

terrorist96:
> otherwise than in pursuit of a Warrant

would become

> otherwise pursuant to a Warrant

or

> otherwise than in pursuance of a Warrant.

Also, the capitalization of Warrant should be standardized.

KacperZajc:
otherwise than pursuant to a Warrant?

terrorist96:
"Than" is unnecessary in that case. At least from my understanding. We need some English majors in here. :laughing:

terrorist96:
I think 'than' is superfluous in all cases. "Otherwise" is correct, or "other than" is correct.

Jean-LouisMesic:
Well done, @terrorist96 ! Good catch.

PR #133, opened 2015-06-14
clean up of warrant language

terrorist96:
Also, I have a problem with the last clause of III.4 saying no person can be detained for more than 24 hours without probable cause. That means that they can be detained for more than 24 hours with probable cause, but no warrant, whereas the first clause says you can't be kept without trial for more than 24 hours without a warrant.


Jean-LouisMesic:
You're reading it wrong.

That provision is about detaining someone while in pursuit of a warrant. For instance, if you catch someone in the act of rape or murder, you need to be able to bring them in before getting a warrant. They can be detained for 24 hours with just probable cause, but after that, a warrant must have been required.

Pursuit of warrant is correct usage in III.4. I've placed the original wording below for comparison's sake.

"* **§III.4.** No person shall be detained without trial for more than twenty-four hours otherwise than in pursuit of a Warrant and unless it is absolutely necessary for the protection of others or where the accused is likely to escape justice or interfere with the investigation; and no detention in pursuit of such Warrant shall be longer than it is absolutely necessary to bring an accused to trial; no person shall be detained for a period not exceeding twenty-four hours otherwise than upon probable cause."

If I am interpreting it correctly, this should be amended, to go with the new warrant language, to be:

* **§III.4.** No person shall be detained without trial for more than twenty-four hours upon PROBABLE CAUSE and in pursuit of a warrant and unless it is absolutely necessary for the protection of others or where the accused is likely to escape justice or interfere with the investigation; and no detention in pursuit of such Warrant shall be longer than it is absolutely necessary to bring an accused to trial; no person shall be detained for a period not exceeding twenty-four hours unless pursuant to a warrant.

terrorist96:
If your interpretation is correct then yes I was reading it wrong. Since they were using the same wording, I didn't think they meant to use it in two different senses (in the process of getting a warrant VS acting in accordance to an already procured warrant)

Jean-LouisMesic:
@terrorist96 Clearly. I'll wait to hear from @KacperZajc to continue discussing this.

KacperZajc:
Sorry Im confused, what is the problem here?
I  can see 'pursuant', yes, we need to change those but what is the issue with the arrest?
Can we make pulls that relate to one change only?

terrorist96:
What part of my PR do you want me to change?

KacperZajc:
make one that changes 'in pursuit of' to 'pursuant to' where applicable;
and explain here what is the problem with the arrest please

Jean-LouisMesic:
@KacperZajc What was your intent from the original section here, as we've found several conflicting interpretations from conflicting wording?

Jean-LouisMesic:
The pulls were relevant to one change, it's just that confusing wording made one section SEEM like it pertained. We need clarification on III.4.

PR #134, opened 2015-06-14
Removed 'any branch of PA', added "any and all"

terrorist96:


terrorist96:
The language about judges being removed by any member of the other branches of the PA leaves open the possibility of a judge being removed by another judge (a member within the same branch of the PA)

Jean-LouisMesic:
@terrorist96 Just something to think about, try leaving typo edits separate from wording changes, and when you're changing words, make a separate pull request for each section, unless you're specifically changing all instances of the exact same word or phrase throughout the document without changing anything else. Makes it easier to merge.

Jean-LouisMesic:
I disagree with this change, this allows agents to act differently than the agency. As long as the "Public X" itself doesn't publicly seize property under IV.13(3)(a), its agents are still able to, in a twisted way. This also brings up the "Should the PA be treated as a person" debate again, to which I sternly decline. Make the agents responsible for the actions of the agency. Leave this degree of separation out, so the "following orders" cliche can be avoided if at all possible. I say we leave "any agent or agency of the Public ___" where ever stated.

Jean-LouisMesic:
@terrorist96 I do agree that the wording of VI.5 should be amended to "any member of any branch of the Public Administration," but I do not agree with your other changes.

terrorist96:
I think it comes down to our different interpretation of an agency and its agents. When I read 'by the public authority,' that it encompases all three branches and all the people in those branches (corporation ≈ people), and that's the difference I guess.

Jean-LouisMesic:
Corporation should NOT equal people, though. This is why we have so many problems in the U.S. Because corporations are valued more than actual persons. Let corporations be people in the countries we leave behind. Let's not bring this horrible idea with us and let it consume our nation. Leave any and all responsibility of the actions of government specifically at the hands who acted.

terrorist96:
I vote that we say people living in Liberland are persons, corporations are people. And then define them separately elsewhere.

Jean-LouisMesic:
What is the benefit of making corporations people?

terrorist96:
Because they are composed of people. People ≠ person. Persons will have the right to vote, people will not.

Also, I proposed "Unconstitutional to make laws for specific entities to give them special protections" so it would tamper any worries you have of corporations getting special treatment.

Jean-LouisMesic:
Not really. Corporations are not people nor persons. They are firms. Firms are not people, they are businesses. Businesses are not people, they are sources of employment and providers of goods and/or services. Naturally, as not people, they should share vaguely similar, very different rights. We need to make this firm and clear in the Constitution, and as soon as possible.

KacperZajc:
judges (other than Justices of the Supreme Court) must be removable by some form of special commission within their independent governing body (this is a matter of public law rather than constitution, my wording just allows this possibility)

PR #135, opened 2015-06-14
Otherwise than vs other than

terrorist96:


Jean-LouisMesic:
@terrorist96  The wording for **§IV.16.** is correct, as it allows the cabinet to revise a rejected bill ONLY by resubmitting it the next year, and by no other means. "Otherwise than by" is the way to say this.

The wording of **§III.10.** is odd, I agree. Deserves discussion. I think your fix is sufficient, but perhaps not.

terrorist96:
I think you're right about IV.16

KacperZajc:
ok make a new pull here

terrorist96:
Is the "fixed IV.16" PR not good?

KacperZajc:
I can’t automatically merge this pull request.

PR #138, opened 2015-06-14
Added taxation to legislative restrictions section

terrorist96:
Also added a period to the end of II.17


Jean-LouisMesic:
@terrorist96 I like this rework. +1

KacperZajc:
we dont touch taxation at this point but if it stays for good, we will get back to your pull :)

terrorist96:
I made a separate PR for the period at the end of II.17

terrorist96:
Please keep this open. Since the tax bill is already in the current draft, this change should be made.

KacperZajc:
but what you want to introduce is already said in any event

terrorist96:
Yeah but the wording is awkward

Jean-LouisMesic:
I like his rewording, it doesn't change anything, but makes it more concise. It also puts the legislative restrictions on taxation where they belong.

PR #140, opened 2015-06-15
shall be construed as to preclude

terrorist96:


Jean-LouisMesic:
Seems good to me, but why is it in its own item in the first place?

PR #142, opened 2015-06-15
treaty ratification

terrorist96:


Jean-LouisMesic:
@terrorist96 I prefer the heavy majority, yes, though is the Cabinet able to introduce this "treaty bill" at any time, and now just at the annual speech?

KacperZajc:
now:
simple majority for singing a treaty by cabinet
but 2/3 for ratification by assembly

are you sure we need have majority for signing as well?

terrorist96:
It says Assembly Representatives, not cabinet

KacperZajc:
it is the cabinet signing the treaty with a simple majority of all assembly representatives 
it is 2/3 of all assembly representatives ratifying the treaty
only then it becomes the law of the land

terrorist96:
What's the point? Can't the cabinet sign anything it wants? Why does it need a simple majority by the Assembly?

KacperZajc:
to make it more difficult to sing international treaties as potentially dangerous for liberty

terrorist96:
But signing is meaningless without 2/3 ratification, right?

KacperZajc:
signing would bind the government vis a vis other states / international organisations (at international level), ratification makes it enforceable in a court of law

terrorist96:
If it isn't enforceable in a court of law, how would the government be bound by it?

KacperZajc:
e.g.: Cabinet signs the European Convention on Human Rights (list of rights plus jurisdiction of international court) - if the treaty is not ratified, citizens could enforce their rights in this international court only and not in domestic courts; if the government breaches some of the rights listed in the convention then citizens cannot enforce their rights in domestic court but the gov is in breach vis a vis the Council of Europe (inter-government organisation managing the Convention and this international court), the organisation can reprimand or expel Liberland etc.

terrorist96:
So then that supports my motion to have 2/3 majority for both signing and ratifying. What if 51% of the assembly agrees, it gets signed, but never ratified?  Then the situation you posed above becomes a reality. 

Also, what can the international court do to liberland? Impose trade sanctions? Because outside of that, I don't see how they could harm the nation for signing but not ratifying something. And if they expel liberland, then so what? We never ratified the document to begin with, so there's really nothing to "expel."

Jean-LouisMesic:
Who is to say that the heavy majority will both agree to sign AND ratify a treaty, however, @terrorist96 , just to play Devil's advocate?

There's no good way to avoid it, but certainly, I believe raising the threshold is warranted. We do not want to be bound by treaties, especially when we have a.) no army to use in accordance with a treaty and b.) no army to use to defend ourselves when we don't provide military support to an ally and they retaliate.

terrorist96:
That could also be a situation, who knows? I'd go so far as to say you need 2/3 in the referendum in addition in order to ratify a treaty.

Jean-LouisMesic:
I'd also be open to that.

KacperZajc:
but then what is the difference between signing by the cabinet with the consent of 2/3 and ratifying by 2/3 in a separate act?

terrorist96:
How about saying that nothing can be signed without being ratified first? Get the 2/3 majority to ratify, then it can be signed.

KacperZajc:
the point is it doesnt make sense to separate signing and ratifying if both require consent of 2/3

terrorist96:
Let's back up and understand what we're trying to accomplish here. We don't want any bad treaties to be enacted without a heavy majority. From my understanding, signing is nothing but a symbolic gesture. Ratifying is what actually matters. But we don't want any signed treaties without ratification either, right? I'm just trying to understand what the ultimate goal is. My suggestion is to have ratification first, then it can be signed. I'll admit I'm not too well versed in treaty laws.

terrorist96:
How about a different solution: 3/5 required for consent, 2/3 for ratification. Thoughts?

KacperZajc:
let's make it 2/3 for signing (which is a first act) and 3/4 for ratifying (which makes it law of the land) :)

PR #157, opened 2015-06-16
Added jurisdiction

terrorist96:
Or would "under the jurisdiction" be better?

### DIFF
--- a/Liberland-constitution.md
+++ b/Liberland-constitution.md
@@ -6,7 +6,7 @@
 
 ## Bill of Rights
 
-The Bill of Rights shall constitute the integral part of the Constitution and shall be binding upon all branches of the Public Administration. The following rights shall apply equally to all persons within the jurisdiction of the Free Republic of Liberland unless stated otherwise.
+The Bill of Rights shall constitute the integral part of the Constitution and shall be binding upon all branches of the Public Administration. The following rights shall apply equally to all persons under the jurisdiction of the Free Republic of Liberland unless stated otherwise.
 
 
 ### Article I: General Provisions
--- a/Liberland-constitution.md
+++ b/Liberland-constitution.md
@@ -6,7 +6,7 @@
 
 ## Bill of Rights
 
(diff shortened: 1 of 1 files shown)


Jean-LouisMesic:
Nay. Foreigners deserve the same rights when dealing with the PA, even when outside of the nation's jurisdiction.

terrorist96:
How can they be guaranteed all of these rights when they are in a different jurisdiction? For example, they can't be afforded the arms rights if they're in a country like North Korea.

Jean-LouisMesic:
Shall we legalize de facto Guantanamo Bay situations for our police forces stationed globally in our embassies? I'm sure we're all just fine with our police from the Liberland Embassy to North Korea in Pyongyang randomly detaining, questioning, torturing, and conscripting North Korean citizens without probable cause or warrant, or convicting them of conspiring to commit terrorist acts without proper trial or due process otherwise.

They are not necessarily under our jurisdiction, but deserve protection from our agents just the same.

Jean-LouisMesic:
They are guaranteed those rights in relation with our government by this document and nothing else, just as the document does not protect you from being publicly shamed for your words despite freedom of speech.

terrorist96:
> Shall we legalize de facto Guantanamo Bay situations for our police forces stationed globally in our embassies?

If you are a Citizen and have a Passport, then Liberland would have technical jurisdiction over you. Foreign embassies would still also be under Liberland jurisdiction.

So maybe 'under' is better than 'within' since the latter insinuates Liberland's geographic limit while the former doesn't.

KacperZajc:
person under jurisdiction is any person within borders of the state, including its embassies
we can change 'within' to 'under'

Jean-LouisMesic:
This still leaves foreigners void of rights, as they aren't necessarily within our borders when our officers deal with them (foreigners aren't allowed in embassies except in special cases.) I'll open a separate issue, however, since this is closed.

PR #158, opened 2015-06-16
Express and informed consent

terrorist96:
Added 'express and informed' prior to most instances of 'consent'. Changed 'is guaranteed'  to 'shall be guaranteed'

Also I don't understand "or with consent or knowledge thereof" regarding cruel punishment.


Jean-LouisMesic:
Yea, good catch here. :+1: 

@terrorist96 That's from a frankly strange debate a few weeks ago about how the Constitution created several strange victimless crimes (outlawing BDSM) without that provision.

terrorist96:
I'd like to change the wording of it, but I'd like to hear from @KacperZajc about the cruel punishment clause.

KacperZajc:
regarding cruel punishment, this phrase was designed to deal wth situations in which an agent of the PA does not torture himself but supervises such practice thereby avoiding any liability

PR #174, opened 2015-06-21
fixed quotes and capitalization

terrorist96:


KacperZajc:
maybe we can change it to "aye" & "nay"?
what do you think?

terrorist96:
Done.

PR #181, opened 2015-06-27
Defining "victim" in victimless crimes section.

yopdog:
I believe this small change will strengthen this section by making it clear that a victim has to be a person and cannot be "society" or any other creative interpretation. Furthermore, it will make animal rights unconstitutional which I believe is the right thing, though some of you may disagree.


KacperZajc:
Im in favour of animals rights.
I understand your concern - let's exclude group of people instead of specifying that it is only a person that can be a victim
so - 

No law shall create victimless and/or consensual crimes, no person of age and/or a group of persons, regardless of its number, shall be considered a victim of one's own actions or omissions.

??

terrorist96:
I'm confused as to why there are so many unrelated commits in your PR.
Regardless, I'm in favor of animal rights too, but only for 'intelligent' animals (dolphins, elephants, whales, monkeys/apes) not stuff like mosquitoes, insects, etc

KacperZajc:
I agree. It's not a matter for the Constitution though. 
But excluding groups of people seems like a good idea to me.

terrorist96:
:+1:

yopdog:
@KacperZajc I find your suggestion a bit confusing. what does it mean that persons cannot be considered victims of one's own actions? Regardless, It doesn't address the change I was trying to make which was to define the word "victim" so this section will have some meaning.

I understand you both support animal rights, but the problem I have with them is that it will make using animals for anything completely illegal. that includes eating them or even running a zoo.

I get it that your interpretation of "animal rights" might have been more akin to partial animal rights, but I believe that puts us in a very arbitrary and inconsistent territory. Why is it ok to torture and kill animals for food but not for fun? If a person is tortured and killed does the reason for the act change the severity of the crime? Furthermore, why is killing animals even allowed in the first place? Isn't the right to life the most basic right of all? 

My belief is that rights are an all or nothing principle. Either they have rights or they don't. Anything in the middle seems to me arbitrary and inconsistent.

yopdog:
@KacperZajc I disagree with you that this is not a matter for the constitution. I can't think of anything more relevant to the constitution then clarifying who falls under the protection of it's rights.

terrorist96:
I think we can have reasonable regulations like you can only kill an intelligent animal humanely, not torture it. And zoos would have to take good care of the animals. I understand your concern regarding the arbitrariness of animal rights. I don't have a good answer to that right now.

KacperZajc:
@yopdog you misunderstood me - Constitution is the place to define who can be a victim, I said it wasnt a place to deal with animal rights but I dont want to exclude them up front

I want to specify that groups cannot be victims as you suggested, I just dont want to mention animals :)

yopdog:
I don't think I misunderstood you, I was actually making the point that animal rights should, in fact, be addressed in the constitiution. To me, defining who can be a victim is basically defining who has rights, which makes the issue of animal rights relevant. 

Nevertheless seeing how the issue of animal rights is controversial, I'm guessing it will not appear in the constitution, so we should at least find  a way to make sure "victim" is not interpreted as "society".

yopdog:
I made a change my PR so it reads the following:

"No law shall create victimless and/or consensual crimes; no person shall be charged with committing a crime against society, nor shall a person of age be considered a victim of one's own actions or omissions."

Any thoughts?

KacperZajc:
"No law shall create victimless and/or consensual crimes; no person shall be prosecuted for any offence against society or any other group, nor shall any Individual be considered a victim of one's own actions or omissions."

yopdog:
I updated it to what you suggested, except I took out "or any other group" since that could include companies which should have the right to sue individuals.

KacperZajc:
other communities?

yopdog:
Communities can still be groups that can represent themselves in court. Why isn't just saying society enough?

KacperZajc:
I just want to make it more demanding. Maybe "society or part thereof"?

yopdog:
What do you think of writing it like this? 
I think that should cover everything. 

"No law shall create victimless and/or consensual crimes; no person shall be prosecuted for any offence without a clear description of the victims and the damages they have incurred, nor shall any Individual be considered a victim of one's own actions or omissions."

KacperZajc:
I like this one: 

"No law shall create victimless and/or consensual crimes; no Person shall be prosecuted for any offence against society or any part thereof, nor shall any Individual be considered a victim of one's own actions or omissions."

yopdog:
Your suggestion is good, but I believe my previous comment's suggestion will cover every possible situation.  
I would like to hear what @terrorist96 thinks about this.

It basically boils down to these two:

"No law shall create victimless and/or consensual crimes; no person shall be prosecuted for any offence without a clear description of the victims and the damages they have incurred, nor shall any Individual be considered a victim of one's own actions or omissions."

"No law shall create victimless and/or consensual crimes; no Person shall be prosecuted for any offence against society or any part thereof, nor shall any Individual be considered a victim of one's own actions or omissions."

terrorist96:
I appreciate the forethought of a situation where someone claims that drug use is illegal because society is the victim of such actions, however I don't think that such a situation would come up. How can society be the victim when we don't have social services (like society paying for the hospital visits of heroin addicts)?

KacperZajc:
Im changing it manually

PR #199, opened 2015-07-03
Safe use of arms

terrorist96:
This would allow for laws against recklessly shooting around.


yopdog:
This could also allow the Goverment to make laws for people's 'own safety'. I don't think this change is necessary since it should already be possible to make laws against putting others in danger.

yopdog:
Maybe you could add something along the lines of: "so long as no negative rights of others are infringed" so it reads the following:

>§II.23. No law shall deprive any Individual of the right to own, manufacture, sell, transfer, transport, bear and use any small arms, arms accessories, or ammunition, be they now historic, contemporary and/or experimental, regardless of condition, so long as no negative rights of others are infringed; no form of licensing and/or registration shall be used to infringe upon these rights; the owner of the premises shall determine whether arms can be carried on such premises except for Agents of the Law Enforcement where acting pursuant to a Warrant.

KacperZajc:
I think this is unnecessary, potential loophole.

terrorist96:
What do you think of @yopdog's wording?

KacperZajc:
I think both versions are unnecessary. 
Unsafe use causing harm is subject to criminal liability, unsafe use not causing harm shouldnt be regulated because the gov shouldnt be able to tell you what is and what isnt dangerous (and this is the case where there is no victim)

terrorist96:
My worry is that someone could try to declare manslaughter unconditional because it violates their right to use arms. Criminal codes need to comply with the Constitution first and foremost.

KacperZajc:
But you punish for the harm, not unsafe use

terrorist96:
So let's forget my 'safety' wording and focus on @yopdog's "negative rights" exemption.

KacperZajc:
and why is it necessary?

yopdog:
It may not be necessary, but you could say the same to the change I made in Section II. 2 which reads:
>nor shall it regulate the use of property, so long as no negative rights of others are infringed

The 'negative rights of others' is the exception for all the freedoms the constitution grants. We should find a why to consistently point that out throughout the constitution.

terrorist96:
Same reason why we have the negative rights exemption for free speech. Here's a better solution: mention the negative rights exemption towards the beginning of the Constitution so as to apply to everything (same way we defined warrants and individuals).

lunokhod:
Consider the following case: Someone is firing a weapon in the air. Most of the time, no one will be harmed, but in the rare cases where harm occurs, it is fatal. This unsafe behaviour thus would invalidate  the premise of first line of the constitution

>  ...in order to secure Liberty, Life and Property...

KacperZajc:
see #204

PR #200, opened 2015-07-03
Section II. 2 adding the word "proximity" and other small changes

yopdog:
Allowing the ownership of all the ground underneath one's land is problematic since it can make building tunnels in Liberland very difficult. Builders would have to ask every land owner for permission to build underneath their land even if it doesn't affect them. 

I therefore suggest we change it to "anything found in proximity under one’s land" and leave it the Assembly to decide exactly how deep underground land ownership extends.

I also made a change to the wording by replacing "and no law shall regulate its use."  with  "nor shall it regulate the use of property so long as no negative rights of others are infringed". I believe this makes it clear that all forms of property are exempt from regulation and not only plants. 

>II.7. No law shall allow for confiscation of private property, either fully or partially, for the public use without the express and informed consent of, and mutually agreeable compensation to, the owner; nor shall it regulate the use of property, so long as no negative rights of others are infringed; anything found in proximity under one’s land or growing naturally on it shall belong to oneself.


KacperZajc:
seems good to me, @terrorist96 what do you think?

terrorist96:
Looks good to me.

PR #203, opened 2015-07-04
Section II. 7 further changes to consider

yopdog:
No section in the constitution prevents the government from transferring property from one person to another. As of now, section II.7 only addresses situations where the government confiscates property for the public good.

The changes I made fix that, but it should be noted that this will also make any cancellation of debt illegal since it entails the transfer of property.     

>II.7. No law shall allow for confiscation of private property, either fully or partially, for any purpose other than reparation or the collection of evidence pursuant to a Warrant; nor shall it regulate the use of property, so long as no negative rights of others are infringed; anything found in proximity under one’s land or growing naturally on it shall belong to oneself.

Any thoughts?


terrorist96:
This is good because it would cover civil asset forfeiture too. But we should keep the existing section too about eminent domain. 

And no 'in' prior to 'pursuant'

yopdog:
ok, I removed the 'in'

yopdog:
doesn't this change cover eminent domain as well?

terrorist96:
Reparations are payments in order to rectify past transgressions. Eminent domain with mutual agreeable compensation doesn't involve any transgressions.

yopdog:
If it's mutually agreed then by definition it's not confiscation, therefore there shouldn't be anything stopping the government from just buying something from someone.

terrorist96:
You have a point there. So what would the reparations provision serve to do? Government taking your property by force in order to give to someone else for being wronged? Sounds like redistribution of wealth.

yopdog:
Reputation is when a criminal pays for the damages he inflicts on his victims. I don't know how you interpreted it to mean 'paying for other people's misfortune'. 

This is at least how I understand it from looking it up on Wikipedia: https://en.wikipedia.org/wiki/Reparation_(legal)

terrorist96:
I guess that makes sense. My interpretation was that the government could take stuff from one person and give it to an unrelated group. As it stands now, I'm unopposed to your changes.

PR #206, opened 2015-07-08
New provision: Determining when a person can be detained or imprisoned

yopdog:
I propose a new provision that will make it so that any form of imprisonment or confinement will be allowed only on the condition that a person can be labeled as a threat to others.  

In the case of detainment, it will ensure that police will conduct arrests only when it's absolutely necessary for the protection of others and their property. If they want to bring someone in for questioning, they can do that provided they have a warrant.

In the case of imprisonment, it will ensure that only people who are actually dangerous to society sit in prison. Putting non-violent offenders in prison is both unnecessary and expensive, I see no reason why that should be the case. 

As of now, the Constitution allows for people to be detained for up to 24 hours even if it's for small misdemeanors and civil cases. 

We should remember that arresting a person involves taking away his liberties for a while, and it should therefore be limited to situations where it is absolutely necessary.

 
>§II.40 . No law shall allow for any Person to be detained or imprisoned, in any way, provided that he poses no future and/or present threat to others and their property, otherwise pursuant to a Warrant.

### DIFF
--- a/Liberland-constitution.md
+++ b/Liberland-constitution.md
@@ -74,6 +74,7 @@ The Bill of Rights shall constitute the integral part of the Constitution and sh
   * **§II.39(6)** to not be compelled to be a witness against oneself, or Persons whose penalization one would rightfully feel as the penalization of oneself;
   * **§II.39(7)** to the assistance of legal counsel free of charge where appropriate;
   * **§II.39(8)** to the assistance of an interpreter if one does not speak the language in which the hearing is conducted, free of charge where appropriate. 
+* * **§II.40.** No law shall allow for any Person to be detained or imprisoned in any way provided that he poses no future and/or present threat to others and their property, otherwise pursuant to a Warrant.  
 
 
 ## Political Institutions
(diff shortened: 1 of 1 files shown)


KacperZajc:
Im afraid this is not a good idea. 
A person should be detained for 24h upon probable cause so that he or she does not escape or destroy potential evidence before the police get a warrant.

yopdog:
Ok, I see your point there. How about we make a small change to provision II.24. by adding "where it is absolutely necessary" in the conditions for detainment? It's pretty vague, but it will ensure that police only arrest people for the reasons you cited in your comment and not for petty crimes where it is unnecessary.

>II.24. No law shall allow for any Person to be detained randomly or without any evidence but only under specified charges, upon probable cause and where it is absolutely necessary with an Agent of the Law Enforcement acting openly and informing the detainee about the rights one enjoys under this Constitution and any laws passed by the Assembly in accordance thereof.

KacperZajc:
Isnt that already guaranteed in §II.25:

No law shall allow for any Person to be detained without trial for more than twenty-four hours otherwise than pursuant to a Warrant and unless it is absolutely necessary for the protection of others or where the accused is likely to escape justice or interfere with the investigation; no detention pursuant to such a Warrant shall be longer than it is absolutely necessary to bring an accused to trial.

???

yopdog:
§II.25 only adresses the situation where a person needs to be detained for MORE than 24 hours. It states that in order to prolong a person's detainment a warrant has to be issued AND it has to be absolutely necessary. Provision II.24. on the other hand, concerns situations occurring before a warrant has been issued and at the time of arrest. 

The change I want to make will ensure that even before the 24th hour, police need to have a very good reason for detaining someone. It can't be just because a person broke a contract or committed a minor misdemeanor, but only when a person is a threat, can escape or destroy evidence.

KacperZajc:
yes, you're right, Im adding "where it is absolutely necessary" to new changes, see #208

PR #217, opened 2015-07-18
Large revision to §II.7 - Private Property Rights

Auahi:
§II.7. No law shall allow for confiscation of private property, either fully or partially, for any purpose other than reparation or the collection of evidence pursuant to a Warrant; nor shall it regulate the use of property with respect to zoning laws; anything found in proximity under one’s land or growing naturally on it shall belong to oneself.
§II.7(1) Any Individual who wishes to challenge the zoning requirements for his/her property should be allowed to do so.
§II.7(2) Any zoning dispute outcome made in court can be challenged by the citizens of Liberland, a referendum shall be held if 3% of the Citizenry object to the ruling. The court ruling will be deemed null if a majority of the Citizenry vote against the ruling. Any referendum held for zoning disputes must begin within 35 days of the court ruling. All construction work on the property shall halt if a referendum based on it's zoning requirements has begun; construction work can begin immediately if a referendum does not change the zoning requirements of one's property.

This will make it so a company can't erect a sky scraper in a residential neighborhood, but still allow for zoning laws to change when the time comes.


terrorist96:
Nay. I cite Houston as example of no zoning laws.

Auahi:
Well how would you like it if you had a very nice neighborhood and a company decides to build a chemical factory in it? This still allows for change...

terrorist96:
If the chemical factory causes pollution or degrades your soil, it's violating your private property rights already. No need for zoning.

Auahi:
Well i just feel that companies shouldn't be allowed to degrade property value... Like by my house they tried to put a solar panel farm which would completely devalue my property and make it hard to sell...

I just wouldn't want things like this to happen.

terrorist96:
Property will always be valuable in Liberland solely due to scarcity. At least that's how I see it.

Auahi:
Yes but would you rather buy the property by the factory or next to the sky scraper or the one next to the river.

And I'm not trying to attack you if it seems like it, sorry, I'm just trying to raise my point...

Auahi:
http://www.factoryfarmtaxprotest.org/examplesofpropertiesdevaluedbyfactoryfarms.pdf

terrorist96:
No, debates like this are necessary in order to form the most perfect Constitution.
I see your point regarding buyer preference. I think that land tax could be used as a form of zoning (for example, charge a higher tax for a commercial property that wants to build in a residential area.

Auahi:
Actually that sounds like a really good idea! I'm still in favor of zoning requirements but maybe something like that could be written into the constitution

Maybe only commercial zones have property tax and the higher the tax based upon where it's located, but the problem with this is it makes it hard for small businesses to find a nice spot within a city. Also this would be awfully reminiscent of taxation through the Bahamas...

terrorist96:
> but the problem with this is it makes it hard for small businesses to find a nice spot within a city

I'd argue that it's good to have some areas with mega businesses and other areas for small businesses. That way, we can have skyscraper districts and local small business districts.

Auahi:
Yeah that would make sense

yopdog:
Absolutely not! Zoning laws are an infringement on property rights   and are exactly what this provision is meant to prevent. Furthermore, they are unnecessary. If a property nearby is polluting in any way, whether it's noise, chemical or danger, it can be sued for damages. 

As for the case of a property degrading the property value of its neighbors, this not a sufficient reason for interfering with one's property since there isn't a direct infringement on other people's rights. Furthermore, it can be argued that any modification of one's property can affect the surrounding property value to some degree. If a person paints his house in an ugly color it might have an effect on property value, but that shouldn't mean he now loses control over his property just because his neighbors think it's ugly. 

That said, there are many ways in which construction can be controlled without the use of laws. My guess is that economics and property rights will be what regulates construction and that would probably work better than any top-down Zoning laws. As @terrorist96 said, Houston can be seen as an example of this.

Auahi:
I like @terrorist96's example of what should be done
 maybe make fluid zoning laws where taxes dictate everything, for example taxes for large business is way more expensive in residential areas than commercial areas

and this edit wouldn't define the zoning laws rather just what to do when there's a disagreement 

and i do admit it would be nice to have a mega business zone, a small business/residential zone, and a separate residential zone

Auahi:
Or at least separate mega businesses from everywhere else and just have two zones: regular and mega-business

terrorist96:
I think this will resolve itself. The mega businesses location will be selected naturally when the first skyscraper is built. The subsequent skyscrapers will pop up around the original.

Auahi:
@terrorist96 I could agree with that

KacperZajc:
Im against zoning. If the factory spoils your soil, this infringes your rights. Freedom to use one's property is subject to NEGATIVE rights of others, i.e. it covers interference with neighbouring land.

Auahi:
Would you consider noise as a interfereance on neighboring land, just curious...

KacperZajc:
depends on the actual loudness, it would be determine by the court :)

Auahi:
Oh okay, i guess not having zoning woulldnt be that bad...

I also think that businesses and skyscrapers would tend to want to stick together anyways, who knows lol...

PR #219, opened 2015-07-18
Small change to citizenship law

Auahi:
§II.31. No law shall enable involuntary loss of citizenship; an Individual wishing to renounce his/her citizenship must do so voluntarily and with intent to relinquish citizenship.
§II.31(1) No law shall prohibit multiple citizenships.

I don't think the state should have the right to revoke someone's citizenship, and I wanted to instate dual citizenship into law.


terrorist96:
Dual citizenship is already implied based on the wording (no person shall be rendered stateless), so only dual citizens are even eligible to be stripped of their citizenship.
Renouncing one's own citizenship is also implied because no laws can be made against it, since it's essentially a victimless act.

Auahi:
I saw it differently, many countries don't allow dual citizenship, but they also don't allow you to become stateless if you want to renounce your citizenship. For example Japan only allows for single citizenship. If you only have Japanese citizenship you can't renounce your citizenship, but inorder for the government to acknolodge that you have lost your Japanese citizenship you need to show them proof you have another citizenship readily available right after you renounce it...

I prefer the way I wrote how you can loose citizenship in this act since it specifically states that you are the only person who can renounce your own citizenship, the previous act stated that the state could do so as well

Auahi:
Here's an example: I have Librish (this sounds way better than Liberlander) citizenship and Polish citizenship, Liberland knows I have both, the previous act stated this:
		-* **§II.31.** No law shall allow for any Citizen to be deprived of one's Citizenship other than by a court of law; no Person shall be rendered stateless in any event.

The "other than by a court of law" means that the government can go to court to revoke my Librish citizenship since I have Polish citizenship, many countries are like this with their citizenship laws and I don't want Liberlands to be like that

Also the voluntarily with intent means a lot since it doesn't allow for a person to do something like run for office in another country and automatically loose citizenship, if some new law stated so

I think Poland has a really good model for how citizenship should work

terrorist96:
I would go one step further and allow multiple citizenships, not just dual. Let's see what @KacperZajc thinks.

Auahi:
Okay I'll change It

KacperZajc:
There is no rule against multiple citizenships. We allow it.
We already guarantee that no one is stateless but we want to keep this power to deprive one of his or her citizenship in some circumstances (to be regulated by ordinary law). 
The current wording seems adequate to me.

Auahi:
I know liberland has said that they allow multiple citizenships but its not instated into law

And still i dont think the state should ever have the right to deprive a person of citizenship even if you have another

What circumstances would you consider to revoke citizenship?

KacperZajc:
The President mentioned to me that we wanted to retain this possibility at this point. 
It will probably be connected to certain criminal convictions.

PR #220, opened 2015-07-18
Huge filibuster loophole fix

Auahi:
* **§III.8(1)** Any Assembly Representative who is found filibustering on duty shall have his/her speech regarding the proposed bill put to a halt.
  * **§III.8(2)** To deem someone as filibustering, the following requirments must be met:
    * **§III.8(2)(a)** The speech must be at least one hour long. 
    * **§III.8(2)(b)** Three fourths of the Assembly Representatives must vote 'Aye' to deem an Assembly Representative of filibustering.
      * **§III.8(2)(b)(A)** A vote regarding an Assembly Representative filibustering shall be held immediately if at least five Assembly Representatives indicate they would like to vote, putting a temporary pause on the speech being voted on; the speech may proceed if no change has occurred.

this was added under: **§III.8.** Every Assembly Representative shall have the right to express one's opinion on the proposed Bill at least once before the vote takes place, without any time limit.

The without any time limit makes it very easy to severely postpone a bill. It's good cause it doesn't set an exact time limit, but it makes so that people who take a huge amount of time that aren't adding any information not able to. aka filibustering

This is worded a little weird but I think it makes sense overall if you have any suggestions do tell

### DIFF
--- a/Liberland-constitution.md
+++ b/Liberland-constitution.md
@@ -127,7 +127,7 @@ The Free Republic of Liberland shall be governed by the Public Administration in
   * **§III.8(1)** Any Assembly Representative who is found filibustering on duty shall have his/her speech regarding the proposed bill put to a halt.
   * **§III.8(2)** To deem someone as filibustering, the following requirments must be met:
     * **§III.8(2)(a)** The speech must be at least one hour long. 
-    * **§III.8(2)(b)** Three fourths of the Assembly Representatives must vote 'Aye' to an Assembly Representative filibustering.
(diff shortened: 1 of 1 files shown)


KacperZajc:
Bills can be submitted once a year only and there is 364 days to pass them and we have only 20 Assembly Reps at this point so there is plenty of time to discuss them.
No need to change filibuster law.

Auahi:
Okay that makes sense i didnt know itd happen only once a year...

PR #228, opened 2015-07-25
Provision §V.4(3) (Arresting of Judges) and changes to §I.10

yopdog:
Provision §V.4(3) cannot be left the way it is since it leaves the police powerless to arrest Judges even in situations where it may be absolutely necessary. There may be cases where there is not enough time to get a warrant from the Supreme Court and damages can be caused because of this. I therefore think we should at least add an exception for 'exigent circumstances' so the police have the power to act in emergencies. 

>§V.4(3) No Judge of any Court of the Free Republic of Liberland shall be arrested for any period whatsoever unless pursuant to the Supreme Warrant **or under exigent circumstances where there is a threat to others or where vital evidence may be destroyed.**

Notice that I took out the part "issued by a simple majority of the Supreme Court" and moved it provision §I.10 with minor rewording. I think it makes more sense to put it there since it fits well with the content of that provision. 

>§I.10. All Warrants in the Free Republic of Liberland, other than the Supreme Warrant, shall be issued by a Judge of the Criminal Court upon request by an Agent of the Law Enforcement; **Supreme Warrants shall be issued by a simple majority of the Supreme Court**; every Warrant shall be supported by evidence of probable cause and shall be as detailed as possible; Warrants granted for a continuous action shall expire no later than after one year and shall be renewed by another Judge in accordance with ordinary procedure.


KacperZajc:
see #243

PR #230, opened 2015-07-26
New provision - The right to collect damages

yopdog:
I think you would agree with me that the right to collect damages is essential for any moral society. As far as I can tell there isn't a provision in the Constitution that clearly protects this right. Additionally, the last part of the provision will prevent the creation of things such as 'Limited Liability Companies' or LLP's. I believe no one should be exempt from liability by hiding behind their company while they are responsible for another person's loss and misfortune.           

>§II.41. No law shall deprive any Individual of the right to collect damages of any kind from those who have harmed him and are responsible for his losses; nor shall it diminish the amount of money paid to a victim to below the cost of damages he has incurred; no special protection against liability shall be created for any individual or group regardless of their position.


KacperZajc:
I cant stress enough how much we NEED limited liability companies!

The right to damages is already guaranteed:
§II.7. No law shall allow for confiscation of private property, either fully or partially, for any purpose other than reparation or the collection of evidence pursuant to a Warrant; nor shall it regulate the use of property; anything found in proximity under one’s land or growing naturally on it shall belong to oneself.

yopdog:
§II.7 doesn't guarantee the right to collect damages, it simply allows for it. Secondly it only barely mentions it in one word. Having a separate provision is necessary if we truly want this right to be protected in the Constitution.    

>I cant stress enough how much we NEED limited liability companies!

Why? don't you think limited liability is unfair? Limited liability basically allows individuals to avoid paying for damages which they are responsible for, thus depriving the victim of the right to collect the full amount in damages he has incurred. I simply see no reason for having it be this way.

KacperZajc:
Limited liability is vital for the economy to function properly. It's out of the question.

We can clarify this bit:

§II.7. No law shall allow for confiscation of private property, either fully or partially, for any purpose other than reparation ordered by a court of law, payment of lawfully levied fines and/or the collection of evidence pursuant to a Warrant; nor shall it regulate the use of property; anything found in proximity under one’s land or growing naturally on it shall belong to oneself.

yopdog:
First of all, let's get one thing out of the way. This provision will NOT in any way prevent the creation of limited liability through contracts. If a company owner wants to be protected from his companies debts he can arrange that with the bank when he makes the contract. 

What this provision will prevent, is being exempt from liability when one has committed a tort. For example, an owner of a chemical factory will have unlimited liability for damages that occur if his factory explodes and injures people around it. I think you would agree with me that it will be extremely unjust to the victims if the owner who is responsible for their loss gets away with his personal assets intact while they need to live the rest of their lives without a leg. 

I don't think giving special protection against liability in such cases is good for the economy. It may be good for businesses owners, but it comes at the direct expense of everybody else. If we have this, we may as well have redistribution of wealth.

KacperZajc:
I dont think this is unfair and even if I did, the whole idea of Liberland is to draw investors so that they can register with our company register and boom the economy. Your idea would have catastrophic consequences, it would kill the whole project. We want legal persons with limited liability, this is exactly what enabled capitalism to flourish.

yopdog:
I thought the whole idea of Liberland was to have a place where human rights are truly respected. Limited liability isn't capitalism it's crony capitalism since it benefits one group of people at the expense of everybody else. Remember, when I'm talking against limited liability I'm talking about torts (personal injury, property damage). I don't have a problem with limited liability being arranged through a contract, that should obviously be allowed.  

I still find it hard to believe you are in disagreement with me, don't you see that limited liability allows people get away with murder?? 

let me lay down my logic:   limited liability = limited responsibility for one's actions  =  limited responsibility for causing injuries = limited responsibility for murder.

I hope you understand now way I couldn't care less about if it makes business less attractive. It is the violation of a very basic moral principle, which is that those who injure or cause damage to another person are responsible for his losses. Waving away this principle for the sake of 'drawing investors' is against everything libertarianism stands for.

terrorist96:
This is similar to the issue I raised in #285

terrorist96:
#235

KacperZajc:
OK I finally see your point. Let me think about that :)

KacperZajc:
see #245

PR #262, opened 2015-08-06

terrorist96:


KacperZajc:
everything is fine except for the torture provision;
we dont want Agents of the PA to perform torture under any circumstances;
if you put 'consent clause' there, it might create a loophole;
this provision does not outlaw torture in private life (like BDSM)

terrorist96:
lol ok

PR #265, opened 2015-08-07
Citizenship can be deprived only by himself

ilyaneba:
That no one would have been able to seize the opportunity to deprive citizenship of its competitor.


yopdog:
>That no one would have been able to seize the opportunity to deprive citizenship of its competitor.

Could you elaborate a bit on the change you want to make? what do you mean by "deprive citizenship of its competitor" ?

ilyaneba:
For examlpe: I'm a Putin :), and I want to exclude Khodorkovskiy from election. Best solution for this deprive his citizenship via my subordinate judge.

KacperZajc:
although it is a valid concern, for now we dont want to close this possibility on direct request of the President

PR #266, opened 2015-08-07
Assembly Representatives election

ilyaneba:
It need to exclude authority monopoly.


terrorist96:
This is already covered.
§III.1(4) No Assembly Representative shall hold the office for more than eight years in total.

ilyaneba:
Yes, you are right!

PR #268, opened 2015-08-08
§I.3 protections for non-citizens/non-residents

Jean-LouisMesic:
Tourists, foreign merchants, and other foreigners were not previously protected under §I.3.


KacperZajc:
let's call it 'and others currently under jurisdiction of the FRL'

Issue #252, opened 2015-08-02
To be dealt with later


ghost:
Here is a list of topics that will be dealt with later:

1. Damages 
2. Electoral system and removal of Assembly Representatives
3. Intellectual property rights
4. International jurisdiction

terrorist96:
Don't forget https://github.com/liberland/constitution/issues/239

yopdog:
You should also add #250: 'privacy and free speech'

KacperZajc:
@yopdog Ive changed it already.

yopdog:
>nothing in this provision shall prevent the collection of information which has been made available to the public or shared utterly voluntarily.

I think the wording can be improved slightly:    
>nothing in this provision shall prevent the collection and sharing of information which has been made available to the public in any way. 

By adding 'in any way' it makes it clear that public information can be shared freely regardless of how it became public. 

Alternatively, we can use this wording:
>nothing in this provision shall prevent the collection and sharing of public information regardless of how it became public.

KacperZajc:
...nothing in this provision shall prevent the collection of information which has been made available to the public in any way or shared utterly voluntarily.

I want to keep 'made available to the public' and 'shared voluntarily' separate. The former refers to information which was published, the latter refers to information which was told the Agents of the PA face to face.

yopdog:
Yes this is good, but what about the 'and sharing' part? I think it makes clearer that the information can be used freely by anyone and not only for LE purposes.

Auahi:
Just had a good idea for taxation and I want to post it here before I forget it, What if we taxed residents of Liberland on income they make in Liberland up to a certain amount. Example:

If I make €100 000 and the average person being taxed makes €50 000 I will only be taxed on what the average person makes so the tax rate will only apply on the first €50 000 I make and I get to keep the rest free of tax, I think a tax rate of 5% to 15% is on income is a good range

Also Land tax is really unnecessary and not right, you shouldn't have to pay to live in your home after you've payed off any loans you have...

KacperZajc:
I really cant stress this enough - there will be ABSOLUTELY no direct taxes

Auahi:
[...]

PR #284, opened 2015-08-11
Provision §II.4. Wording

yopdog:
>nothing in this provision shall prevent the collection of information which has been made available to the public or shared utterly voluntarily.

Changed to:
>nothing in this provision shall prevent the collection and sharing of information which has been made available to the public in any way or shared utterly voluntarily.

??


terrorist96:
Facially, I'm not opposed to this change. However, I have to ask: what purpose do you hope to fulfil by making this change?

yopdog:
'in any way' - makes it clear that regardless of how information becomes public it is free to use by anyone. 

'and sharing' - I think it's just better wording. makes it clear that you can use the information however you want. 

see #252

terrorist96:
> 'in any way' - makes it clear that regardless of how information becomes public it is free to use by anyone.

There's no qualifiers listed for how it became public, so it covers any situation by default. 

> 'and sharing' - I think it's just better wording. Makes it clear that you can use the information however you want.

If you can collect it, sharing is moot. Sharing would only be disallowed for nonpublic information. 

Like I said, I'm not against these revisions, but I think the current language covers the situations you wish to protect.

yopdog:
Yes, I see what you mean, but I still stand on this change. It makes it harder to interpret it differently. What you may see as the default by your logic, may not be what others see according to their interpretation.

KacperZajc:
Im with @terrorist96 on that.

Issue #281, opened 2015-08-09
up for discussion 


ghost:
1. Citizens' right to repeal legislation
2. Minority's right to veto bills
3. The right not to appear in court
4. Judicial election.
5. Religious neutrality
6. Expiration of 'Special Powers' Clause
7. Supreme Court vote 
8. Expiration of laws
9. Polycentric republic philosophy 
10. Jury acquittal expressly based on morality
11. Fair Representation

terrorist96:
> Assembly one-off projects

Private contractors are sufficient.

> Ban/restriction on possession of shares in companies by Members of the PA

The language needs to be crafted very carefully. Also, I'm in favor of requiring candidates/members to disclose the identity of large donors to their campaigns.

> Recall process for Assembly Representatives

Yes plz. And perhaps - *maybe* - for the Chancellor too?

jasonmullins:
I very much favor specifying the right of citizens' to repeal acts.

terrorist96:
> Citizens' right to repeal legislation

Jury nullification is already allowed. 

> Minority's right to veto bills

It'd be nice in concept, but I think it'd cripple the system too much to the point that nothing would ever get accomplished.

jasonmullins:
> Citizens' right to repeal legislation
I don't think jury nullification is the same thing as repealing a bad law.  Are you saying a single jury can decide on nullification and then nobody else will ever be charged using the same law?  In modern government, it is nearly impossible to get rid of an old law because of the vocal minority and apathy.  I just think we need a way to repeal bad laws that doesn't involve elected politicians having to take on the political risk to do it.
> Minority's right to veto bills
I disagree.  I don't think just laws will even make it to this point in the process.  Also, it's not a stretch to think that if you can't get 3/5 to support a law then it's probably an unjust law.  A good law should be non-controversial imo.

terrorist96:
> Are you saying a single jury can decide on nullification and then nobody else will ever be charged using the same law?

No, but a prosecutor would have trouble obtaining a conviction if no Jury would be willing to convict under that law. The law would naturally be repealed itself.
I'm not opposed to allowing referendums for repealing laws; I just don't think it's super necessary because of all the other checks and balances we have in place.

jasonmullins:
That's a good point.  It might not be the most pressing issue - I've just always found it frustrating that there's no good way to get rid of laws.  Also - even if they can't get a conviction, a not-so-noble government could still use the law as a type of harassment as a way to get warrants and stuff.

[...]

PR #286, opened 2015-08-12
Allow minority to veto ordinary bills

jasonmullins:
I believe that a significant minority should be empowered to veto ordinary bills.  This will also help to dampen 2 party politics.  Any thoughts?


KacperZajc:
Im putting it up for discussion #281

PR #288, opened 2015-08-12
General repeal

jasonmullins:
I noticed some debate about the sunset clause.  Something like this could be a compromise.


KacperZajc:
Im putting it up for discussion #281

PR #292, opened 2015-08-14
Public Prosecutor

yopdog:
>The criminal offences in the Free Republic of Liberland shall be prosecuted by the Office of the Public Prosecutor of the Free Republic of Liberland, headed by the Chief Public Prosecutor.

Does the public prosecutor really need to be involved in criminal offences between citizens? Can't this be handled through private prosecution? I see no reason for why we need the public prosecutor to prosecute on behalf of the citizens. After all, when a person commits an offence it should be up to the victim to decide whether or not to press charges. What's more this could potentially act as a loophole for the victimless crimes provision if the public prosecutor chooses to prosecute on behalf of the "victim" even if the "victim" has no intention of pressing charges himself.

Unless there is a good reason for this, I would suggest the following change:       

>§V.10. The Office of the Public Prosecutor shall be the main body in charge of prosecuting on behalf of the Public Administration, headed by the Chief Public Prosecutor.

>   §V.10(1) The Chief Public Prosecutor shall be appointed by the Chancellor for a term not exceeding six years.
   §V.10(2) The Chief Public Prosecutor shall be removable from the office by the Citizens with a majority of two-thirds vote in a referendum called by 10% of all Citizens.
   §V.10(3) The Public Prosecutor shall deal only on issues involving the Public Administration.


terrorist96:
> After all, when a person commits an offence it should be up to the victim to decide whether or not to press charges.

This is already the case.

> -  * **§V.10(3)** Prosecutions shall be brought on behalf of the Citizens or with the express and informed consent of the victim where appropriate.

edwardtjgerety:
This I am 100% agreement.  The governments in the US and Australia use judicial tricks file charges without a victim to press for bigger gains against citizens.

yopdog:
@terrorist96 It says "where appropriate" so there can still be a situation where the public prosecutor chooses to prosecute when the victim has no intention of doing so. Regardless, I still see no reason for the government to be involved in this in the first place.

KacperZajc:
'where appropriate' refers situations where the victim cannot consent (e.g. murder) or there is no one specific victim (e.g. cruelty to animals)

yopdog:
@KacperZajc ok, but I still don't see why the government needs to be involved in this at all. The government should only be involved in things it absolutely has to. Prosecution is something that can handled privately, there is simply no reason this.

terrorist96:
I would support removing `or` from the clause. 
> §V.10(3) Prosecutions shall be brought on behalf of the Citizens or with the express and informed consent of the victim where appropriate.
@yopdog maybe the family of a murder victim doesn't want to burden themselves with a lawsuit in order to make sure the offender gets punished. They always still have the opportunity to sue in Civil Court, right? Or would that constitute double jeopardy? In the USA, it doesn't.

KacperZajc:
We cant remove OR because as I pointed out, there might be no victim to consent in some situations.

yopdog:
@terrorist96 Even in that case there are free market solutions.

see:https://en.wikipedia.org/wiki/Prosecution_association

A bit of history on private prosecution England:

>In the early history of England, the victim of a crime and his family had the right to hire a private lawyer to prosecute criminal charges against the person alleged to have injured the victim.[14] In the 18th century, prosecution was private of almost all criminal offences against the person, usually by the victim.[15] One reason for this was that prosecution had never been conceived of as a public matter. The English system was based on the principles of individual and local prosecution in which the right of the private citizen was paramount.[16][17] The exception to this norm was in offences where the victim was the Crown, and the Attorney-General and the Solicitor-General represented the prosecution at State trials; for an example, see the 1637 ship money tax protest case of John Hampden. https://en.wikipedia.org/wiki/Private_prosecution#England_and_Wales

terrorist96:
@KacperZajc The `where appropriate` part covers that. 
@yopdog That's interesting, but what if they can't afford to hire a lawyer? The Constitution allows for free legal representation in cases where you are the defendant, but not if you're the plaintiff.

yopdog:
@terrorist96 It's my understanding that the budget will be financed voluntarily. Given that fact, what difference would it make if financial help comes from the PA or a private organization? In both cases, the money for 'free legal representation' can be considered a volentary contribution. Furthermore, there are many market solutions for this problem, which could include taking insurance before to cover such costs or taking out a loan that could later be paid back with money collected for damages.

terrorist96:
On one hand, I understand your free-market arguments, but on the other hand, I don't like the idea of Justice being based on your capability to seek it. Following this same argument, we would have to remove free legal representation/interpretation too.

yopdog:
As I pointed out in both cases the money will come from voluntary contributions. Whether that money comes from the government or a private institution makes no difference as to the availability of funds to help the poor.

People with money will always have the better of everything. That said, in the case of justice the victim at least has the right to be paid for damages which include all the legal costs he has incurred. This makes it possible for anyone to afforded justice if they are truly a victim of a crime. 

>Following this same argument, we would have to remove free legal representation/interpretation too.

The constitution grants the right to free legal counsel only to **defendants under prosecution** meaning that if for any reason there are no funds available for supplying free legal counsel to a person, the case will have to be dropped and the person will be free to go. Therefore, if the prosecutor really wants to continue the case against that person he will have to pay for his legal costs himself.

KacperZajc:
prosecution of criminals seems to be the very basic function of a state, arguably even rationale for its existence

terrorist96:
@KacperZajc keeping 'or' in the clause makes it meaningless. If a victim objects to the prosecution, the state can still prosecute because it can claim to be bringing the prosecution on behalf of the citizens in general.

yopdog:
@KacperZajc can you please keep this open? This discussion is in no way finished.

terrorist96:
:+1:

yopdog:
@KacperZajc If you have a good reason for why we need the government to be involved in prosecution I would like to hear it.

KacperZajc:
Because ordinary people have neither time nor means to do that.
But for those who do, the right to private prosecution is guaranteed.
If you want to expand this right, feel free to propose an amendment. 
But public prosecution as such remains, as this is one of the reasons we have states in the first place even under libertarian ideology.

terrorist96:
What about my concerns with keeping the word 'or' in the provision?

yopdog:
@KacperZajc Thanks for being clear on why you are opposed to this.
>Because ordinary people have neither **time** nor **means** to do that.

**Time**: Who said people have to do everything themselves? This is not a good argument since in the free market if you don't have time to do something you just pay someone else to do it for you. You could just as well say the government must grow tomatoes since nobody has time for it. I think you see my point. 
See: https://en.wikipedia.org/wiki/Prosecution_association

**Means**: This is not a good argument either. In Liberland, the budget will be entirely funded by voluntary contributions meaning that the government will not be in any more of a position to help the poor than any other private charity. In other words, having the government provide prosecution for the public will not guarantee prosecution for the poor any more than a private charity, since in both cases the funds are collected voluntarily. Please don't make the mistake of thinking that burdens on the public can be diminished through the government.   

That said, here are a few reasons why I believe we should not have the public prosecutor involved in offences between citizens:

1) The free market does it better - The government should not be involved in anything that could easily be done privately. Private prosecution is something that can definitely be provided through the market and could probably be provided more efficiently and with better quality. 

2) Dependency on government - The more we have to depend on the government to provide us with essential services the more power it will have over us. Why should we depend on the government to prosecute on our behalf when it could just as easily be provided through the market?     

3) It limits the power of government -  Taking away from the government the power to prosecute citizens is a good way to prevent victimless crimes from emerging. If the burden of prosecuting for victimless crimes falls on private individuals it is probably safe to guess they would never be enforced.

KacperZajc:
#305

PR #295, opened 2015-08-16
`may` vs `might`

terrorist96:
I feel like `may` is a better word to use in this context. It allows for discretion. Think of `shall issue` and `may issue` laws in the United States for permits to carry weapons. In addition, I feel like `might` is less formal.


KacperZajc:
Im fine with not using 'shall' here but Ive always thought might is a more formal version.

terrorist96:
Yes, we don't want to use `shall`, because that doesn't allow for discretion. The opposite of shall is may. I've never seen `might` used in a legal context.

PR #303, opened 2015-08-21

terrorist96:


terrorist96:
Actually, this can probably be ignored if we're restating article III entirely.

PR #319, opened 2015-08-26
Important Bill of Rights addition!

Auahi:
So this might sound arbitrary but this is very important, an example would be this: 

If a majority of people in Liberland wanted Sharia Law they "could" pass it since people claim sharia does not promote islam... There are many people who think a secular state can still be secular and have sharia law, well this law would stop or at least make it a lot harder to do so...


Auahi:
Maybe I'm just being ridiculous...

terrorist96:
:-1:

Auahi:
+* **§II.44.** No content be that religous or nonreligous which comes from or has any bases from a religous source shall not be inaugurated into law.

It's not bad it would just be a precaution, but I am being a little paranoid

PR #326, opened 2015-08-28
Cruel or unusual punishment

Auahi:
The prior paragraph just said no law can promote the application of torture or other cruel, unusual or degrading treatment, but this also makes it so courts can't give these as sentencings, since these sentencings aren't considered 'laws', if you understand what I'm trying to get at.

I'm just readdressing the outlaw of this in court.


Auahi:
Actually it's a little unnecessary after rereading.

Issue #305, opened 2015-08-22
Private prosecution 


KacperZajc:
§V.10. The Victim(s) of a criminal offence shall have the inherent right to institute the Private Prosecution of alleged offenders. 
§V.10(1) The right shall be exercisable before the Office of the Public Prosecution institutes any Public Prosecution.
§V.10(2) The Victim(s) may expressly waive the right to institute the Private Prosecution.
§V.10(3) The Victim(s) shall be deemed to have waived the right to institute the Private Prosecution should it not be brought within three months since an alleged offence took place.
§V.10(4) The Private Prosecution may be instituted by the Victim(s) in person or by Proxy on its behalf.
§V.10(5) Both the Victim(s) or the Proxy shall not enjoy the privileges of the Law Enforcement by virtue of pursuing prosecution.
§V.10(6) Should some, but not all of the, Victims waived their right to institute the Private Prosecution, the accused shall be prosecuted by the Office of the Public Prosecutor in cooperation with the Victims willing to participate in the prosecution. 

§V.11. The Public Prosecution of alleged offenders shall be instituted by the Office of the Public Prosecutor.
§V.11(1) The Office of the Public Prosecutor shall retain the right to institute the Public Prosecution should the Victim(s) waive, either expressly or impliedly, the right to the Private Prosecution.
§V.11(2) Should a criminal offence leave no Victim(s) capable of instituting the Private Prosecution, the Office of the Public Prosecutor shall have the inherent right to institute the Public Prosecution on behalf of the state of the Free Republic of Liberland.
§V.11(3) Should there be a Victim capable of consenting, the Public Prosecution shall be instituted with the express and informed consent of the Victim and on its behalf. 
§V.11(4) Should the right to the Private Prosecution be waived and evidence be sufficient, the Office of the Public Prosecutor shall be under obligation to institute the Public Prosecution.
§V.11(5) Should the Public Prosecution be instituted, the Victim(s) shall have the right:
§V.11(5)(a) to be notified about the schedule of the judicial proceedings; 
§V.11(5)(b) to address the Criminal Court before the sentence is determined; 
§V.11(5)(c) to be advised of release from custody or escape of the accused offender.

§V.12. The Office of the Chief Public Prosecutor shall be independent and shall be headed by the Chief Public Prosecutor.
§V.12(1) The Chief Public Prosecutor shall be appointed by the Chancellor for a term not exceeding six years.
[...]

PR #329, opened 2015-08-29

terrorist96:
Removed language from III.22(1) because it's a mandatory referendum, not one that is triggered by the request of the citizens and moved it to III.21(1).
Renumbered IV.6(#)
And other stuff I don't feel like listing.


KacperZajc:
I thought it was 'the Private/Public Prosecution', @yopdog suggested 'a' in #305 
Thanks for being thorough.

terrorist96:
It was @Auahi who wanted it changed to `a`

PR #72, opened 2015-05-09
fully-informed juries

FellowTraveler:
This PR continues the discussion from https://github.com/liberland/constitution/pull/55

PR #55, opened 2015-05-08
Fully-informed Jury nullification.

FellowTraveler:
Added the right of the Jury to nullify the law. See Indiana state constitution. Also added right of Jury to be informed of its power to nullify the law.

PR #344, opened 2015-09-04
Jury has right to judge the law based on personal sense of justice.

FellowTraveler:
I used _"personal sense of justice"_ because the same phrase is already used elsewhere by Liberland law.

This PR continues the discussion from PRs #55, #72, and most recently, #80.

Judges often instruct jurors to do things that violate the jurors' conscience. Many unjust convictions are obtained this way in the United States. Jurors often later express regret at these injustices and their part in them.

Jurors have the right to judge the law (this was the entire historical purpose of the jury) but often courts mis-advise jurors about this fact, and they also excuse jurors who are aware of their right to nullify bad law. That's why it's so important that the entire jury always be clearly informed of their right to judge an unjust law.

The only concern about this previously was that jurors may abuse this process. But in reality, it's the courts who abuse this process, whereas juries only historically use it for good. There was also a concern about biased jurors (the example raised previously was a muslim juror faced with a jihadi defendant.) **But since the entire jury must vote to acquit, this will not be a problem in practice.**

Signed-off-by: FellowTraveler F3llowTraveler@gmail.com

### DIFF
--- a/Liberland-constitution.md
+++ b/Liberland-constitution.md
@@ -360,7 +360,7 @@ The Free Republic of Liberland shall be governed by the Public Administration in
   * **§IV.15(1)** The Jury shall be composed of twelve impartial Citizens drawn randomly from the Electoral Register.
   * **§IV.15(2)** The Jury shall determine the facts and render the verdict according to the law under the direction of the Judge.
   * **§IV.15(3)** The Jury shall deliberate in camera and shall not be required to disclose reasons behind the verdict or be reprimanded for it.
-  * **§IV.15(4)** The Jury shall retain the unequivocal right to acquit and shall be informed of this right; such acquittal shall be final.
+  * **§IV.15(4)** The Jury shall retain the unequivocal right to acquit based on personal sense of justice, and shall be informed of this right; such acquittal shall be final.
   * **§IV.15(5)** The Jury shall convict with no fewer than eleven votes.
   * **§IV.15(6)** All defendants who have been convicted by the Jury shall be sentenced by the Judge as prescribed by law.
   * **§IV.15(7)** The Jury shall render the verdict free from any form of coercion; the Judge may order sequestration of the Jurors should it be required for the independence of the Jury.


FellowTraveler:
Suggested Juror's Oath (I can submit a separate PR for this):

_You have an unequivocal right to acquit based on your own sense of justice, and it is your duty to do so, even if the defendant has broken the law, if you believe that the law itself is wrong, or that the defendant is innocent of wrongdoing. Do you understand these rights, and do you swear to condemn no person who is innocent, and acquit no person who is guilty, and to render a true verdict according to your conscience?_

----------------------------

Because I'm sure we can acknowledge, that if all 12 members of the Jury, chosen at random, all **agree** that the law _itself_ is unjust, then they should have no obligation to convict against their own conscience, which would make them **accomplices** in the injustice perpetrated by the government in the event of an unjust conviction. Certainly it is in the interests of justice that jurors should have the last word on unjust legislation.

The [sordid history](https://en.wikipedia.org/wiki/Jury_nullification) of the American court system proves that otherwise, authorities will chip away at the power of the jury, through precedent and misinformation, until they are able to impose outright tyranny upon the people.

Only by explicitly requiring that jurors be informed of their right to conscience, and by requiring that they must swear to uphold justice above all else, can we protect the citizens and residents of Liberland from the inevitable destruction of their liberties that is currently ravaging America.

terrorist96:
The issue is that as it currently stands, all that is required for an acquittal is for 2 jurors to refuse to convict, not all 12 to do so. Conviction us obtained with 11 or 12 jurors in agreement. Anything less results in acquittal.

FellowTraveler:
Shouldn't that be a hung jury?

Guilt or Innocence should be decided by Unanimity, and otherwise the defendant should face a retrial.

I think that any decisions regarding the Jury need to take into consideration the arguments presented by Spooner since his logic is undeniable. For example, Liberland does not provide the right to a Jury in a civil case, only a criminal one.

But consider Spooner's argument against doing this:

```
Furthermore, it would be absurd and inconsistent to make a jury paramount to legislation in criminal suits, and subordinate to it in civil suits; because an individual, by resisting the execution of a civil judgment, founded upon an unjust [114] law, could give rise to a criminal suit, in which the jury would be bound to hold the same law invalid. So that, if an unjust law were binding upon a jury in civil suits, a defendant, by resisting the execution of the judgment, could, in effect, convert the civil action into a criminal one, in which the jury would be paramount to the same legislation, to which, in the civil suit, they were subordinate. In other words, in the criminal suit, the jury would be obliged to justify the defendant in resisting a law, which, in the civil suit, they had said he was bound to submit to.
To make this point plain to the most common mind—suppose a law be enacted that the property of A shall be given to B. B brings a civil action to obtain possession of it. If the jury, in this civil suit, are bound to hold the law obligatory, they render a judgment in favor of B, that he be put in possession of the property; thereby declaring that A is bound to submit to a law depriving him of his property. But when the execution of that judgment comes to be attempted—that is, when the sheriff comes to take the property for the purpose of delivering it to B—A acting, as he has a natural right to do, in defence of his property, resists and kills the sheriff. He is thereupon indicted for murder. On this trial his plea is, that in killing the sheriff, he was simply exercising his natural right of defending his property against an unjust law. The jury, not being bound, in a criminal case, by the authority of an unjust law, judge the act on its merits, and acquit the defendant—thus declaring that he was not bound to submit to the same law which the jury, in the civil suit, had, by their judgment, declared that he was bound to submit to. Here is a contradiction between the two judgments. In the civil suit, the law is declared to be obligatory upon A; in the criminal suit, the same law is declared to be of no obligation.[115]
It would be a solecism and absurdity in government to allow such consequences as these.
```

We ought also to consider what he says about the Moral Considerations for Jurors:
http://oll.libertyfund.org/titles/2181#lf1474_label_163
_Of course, no man can rightfully take an oath as juror, to try a case “according to law,” (if by law be meant anything other than his own ideas of justice,) nor “according to the law and the evidence, as they shall be given him.” Nor can he rightfully take an oath even to try a case “according to the evidence,” because in all cases he may have good reason to believe that a party has been unable to produce all the evidence legitimately entitled to be received. The only oath which it would seem that a man can rightfully take as juror, in either a civil or criminal case, is, that he “will try the case according to his conscience.” Of course, the form may admit of variation, but this should be the substance._

FellowTraveler:
These articles also are relevant and enlightening:

http://oll.libertyfund.org/titles/2181#lf1474_label_181

http://oll.libertyfund.org/titles/2181#lf1474_label_025

terrorist96:
> Shouldn't that be a hung jury?

I suppose, but we accept innocence until proven guilty, and if you're failed to be found guilty by the jury, you're acquitted because a retrial could result in double jeopardy. 

Appreciate the Lysander Spooner excerpts. Interested in what @KacperZajc thinks too.

KacperZajc:
1) the wording proposed is a terrible idea as it makes any law redundant and passing of any law meaningless (jury should be informed that conviction should be made according to the law but also that nevertheless they have the unconditional right to acquit)
2) the idea of hung jury should be explored (thank you for brining it to my attention)
3) no jury in civil cases (I have already stated the reasons on several occasions)

FellowTraveler:
@KacperZajc 
```
1) the wording proposed is a terrible idea as it makes any law redundant and passing of any law meaningless (jury should be informed that conviction should be made according to the law but also that nevertheless they have the unconditional right to acquit)
```

It certainly does not make any law redundant! The Jury does _not_ have the power of legislation. The laws are written by the lawmakers, and it is _only_ by those laws that anyone will be charged in court. And thus no one will _ever_ be convicted except according to those laws! **So the rule of the majority is clearly in effect via the legislature.**

However, the Jury _does_ have a _veto_ on those laws, as the Jury represents the whole people, thus providing the minority with a failsafe against tyranny.

And before someone complains that the minority will rule the majority, remember that the minority does not legislate, so they cannot enforce their will upon the majority. And isn't it also in the majority's interest that **justice** should be established? Different groups of people form different minorities on each issue, and thus _all_ people are in one minority or another, depending on the case. _All_ share an interest in seeing justice done, especially the weaker party.

```
2) the idea of hung jury should be explored (thank you for brining it to my attention)
```

What kind of kangaroo court are we running here?

And therefore, since all jurors must agree in order to render a verdict (or at least 11 of them, if you want to ward against juror interference) then you don't have to worry about some crazy juror messing up the justice process. Worst case, there will be a retrial, and then the defendant will receive his verdict, without any violation of double-jeopardy. This is normal in common law.

```
3) no jury in civil cases (I have already stated the reasons on several occasions)
```

I suggest you at least answer Spooner's argument (that I pasted above) since he proves this to be a contradiction.

PR #352, opened 2015-10-16

terrorist96:


terrorist96:
Also, which do you want to standardize?
testimonies given in breach of oath shall be prosecuted in a regular manner in the Criminal Court.
or
testimonies in breach of the oath shall be prosecuted in the Criminal Court in a regular manner.

???

terrorist96:
I recommend the first one. What do you think?

terrorist96:
See #354

PR #362, opened 2015-11-01

terrorist96:
Removed 'criminal' so as to have it apply to all possible offences.


terrorist96:
Looking through the Constitution, I don't see references to any type of offence other than criminal. What about civil offenses?

KacperZajc:
we refer to 'civil wrongs' in the Civil Court section
Im not sure about this change tbh

terrorist96:
My reasoning is that I want this to apply to any and all offences which carry some sort of punishment. For example, running a red light probably isn't a criminal offence, so a cop could claim it and they could win a conviction based solely on their word because I.29 applies only to criminal offences.

KacperZajc:
So far we have not referred to traffic regulations at all because it might be the case all roads will be private therefore no universal traffic laws are needed. I think we should stick tot his policy at the moment.

terrorist96:
Okay, outside of that, the issue is still there though. Not every offence will be a criminal offence. If we call it a 'civil wrong', then we should expand this (and all other protections for defendants) to encompass criminal offences **and** civil wrongs. What do you think?

KacperZajc:
Honestly, I dont think so. Civil law governs the relationship between individuals with the state providing forum for resolving dispute only. I dont see what constitutional rights should be conferred upon one individual against another in this context.

terrorist96:
I don't disagree. But are we going to only have criminal offences and nothing else? What about littering on public land? Will that be a criminal offence?

KacperZajc:
Fines are allowed (see §I.44. and §I.60.). And those are decisions of the PA so automatically subject to the right of judicial review (§I.8.). I dont think we need any further protection here.

terrorist96:
So a person could be fined without convicted of a crime? My impression was that those fines referenced fines imposed as a penalty for being convicted of a crime.

KacperZajc:
Hmm we should think of more rigid regulations regarding fines in that case. I reckon we should allow normal fines (without any judicial approval unless challenged before the Civil Court). But maybe we should put some additional safeguards?

terrorist96:
I'm against any fines without court approval. You could make it optional for the person to come to court (like minor infractions), but a court date should always be required.

PR #368, opened 2015-12-01
Small fixes

terrorist96:
Don't merge yet. I still need to go through most of the changes.


terrorist96:
Ready to be merged.

KacperZajc:
the problem with citizenship is that we dont want to give it away to anyone born to a citizen; it would be better to retain this as a prerogative; that's why I put automatic entitlement only where otherwise such a person would be stateless

terrorist96:
That would require the Supreme Court to interpret the constitution and laws of every other country in the world in order to determine if someone is eligible for that citizenship.

terrorist96:
@KacperZajc do you see my point?

KacperZajc:
yes, that's a good point

how about:
"should a Person be born to parents of whom both are Citizens of the Free Republic of Liberland, he or she shall automatically be entitled to citizenship of the Free Republic of Liberland"

???

terrorist96:
Ready to be merged. :)

PR #390, opened 2016-02-08
I removed some ambiguity from §I.34.

Auahi:
"nor shall Minors and those who lack mental capacity be subjected to any permanent and/or harmful treatment which is not medically necessary **for sustaining life**."

_I added the bold_


Auahi:
I need to reword this

PR #394, opened 2016-02-19
removed the 5% land tax

thetimpotter:
If an individual can not refuse service, there is no service.

As seen in the United States, the smallest government ends in the fiercest tyranny.
As seen in ancient Iceland, the temple fee results in concentrated power and turns a free society to a violent collapse.

For these reasons the land fee was removed, and 100% of government activity is funded through voluntary donation or purchases of services.


KacperZajc:
it was already discussed here many time, land fee stays

PR #401, opened 2016-04-08
Untie Assembly salary from spending

YallyWally:
If assembly salaries are tied to spending levels, it will give assembly members a perverse incentive to increase spending.

### DIFF
--- a/Liberland-constitution.md
+++ b/Liberland-constitution.md
@@ -45,7 +45,7 @@ The Free Republic of Liberland shall be governed by the Public Administration in
   * **§II.1(3)** The Assembly may reconvene at any time whatsoever upon the call of the Assembly Speaker.
   * **§II.1(4)** The Assembly shall not be dissolved, nor shall its session be interrupted, by any other body of the Public Administration.
 * **§II.2.** The Assembly shall consist of twenty Assembly Representatives.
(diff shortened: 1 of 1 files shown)


KacperZajc:
That is a good point but we would need some other safeguard instead.
Any ideas?

YallyWally:
I have it tied to GDP in my pull request.

Though I wonder if we even need to compensate Assembly members in the first place. What's the point of paying them if it's just going to be a "token" amount anyway?

terrorist96:
How about we just define a hard number for their pay in the constitution? That's how New Hampshire's constitution is layed out.

KacperZajc:
I am against hard numbers in the Constitution. To much changes overtime.

YallyWally:
Why don't we just make Assembly members unpaid? That would eliminate all of these problems. It's pretty clear that the constitution is laid out in such a way that assembly members will need to maintain alternative sources of income anyway.

HonzaMikula:
BTW from what source of money the salary would be paid? From voluntary taxes, right? So maybe people should voluntarily contribute to salary of Assembly members...

Auahi:
I'm against the GDP idea, it's mainly an estimate whereas the annual budget is a definite number.

terrorist96:
Also, tying it to the annual budget won't necessarily incentivize more spending, since deficits will have to be covered by the legislators.

PR #410, opened 2016-04-27
Removal of ARs

Auahi:
In Liberland an AR needs only 5% of votes to gain a seat in the Assembly, how could we make it possible for population to also allow for removal of this person. What I see happening is people easily being forced out of office by the majority and having the majority political party just take their empty seat. This severely impedes Liberland from becoming a democracy.


Auahi:
An example: there are three main parties in Liberland, A, B and C. A is 55% of the population. B 30%, and C 15%. with how the model works for AR removal every B and C seat on the assembly could be removed by the public and replaced with an A seat since A is the majority.

terrorist96:
I think you're misunderstanding the provisions. You can recall an AR with a simply majority (50% + 1), but the election can only occur if 5% of the voters sign a petition demanding the recall election. How an AR gets elected is still unknown at this point.

KacperZajc:
@terrorist96 is right.

Issue #413, opened 2016-05-02
Land fee discrepancies 


Auahi:
§II.18(6) The overall Fee revenue shall be calculated not to exceed 5% of the Gross Domestic Product of the Free Republic of Liberland of the preceding year.

GDP isn't the best way to calculate taxes, there are so many factors that can change GDP, it doesn't accurately portray wealth. GDP varies widely depending on the source used to calculate it as well. 

Does anyone have any ideas for a new way to calculate the land fee?

KacperZajc:
I agree that GDP is not ideal but as far as I know there is no better yardstick.
I would be happy to hear your ideas.

bjoelf:
How is GDP to be calculated if taxes are voulantary and no public administration to gather data and to do the acctual calculation? In Sweden there is a separate autority to just that (among other things). https://en.wikipedia.org/wiki/Statistics_Sweden.

Just my two cents. :) /Björn

mherzl:
I suggest we get rid of the land fee and require gov to not violate the rights its supposed to be protecting.

Ideas for how to pursue revenue without violating rights include court/citizenship/voting fees, the canton system suggested in other posts, and voluntary taxation.

bjoelf:
The state needs continius and predictible income so in my opinion the Land Fee is not the problem. It's how to create a rational method of calculating it's size that is the core of my consern. Anyone suggesting land fee is not a good alternative can come up with a better proposal. It's voluentary to buy land so charging a yearly fee is not to violate rights it's about to organise a country.  Just another two cents :) /B

Auahi:
@mherzl Im not overly fond of the land fee but I understand why they have it. I'd much rather have a land fee than voting fees though; that could go wrong so many ways. I suggest you make another thread if you want to propose a new system, this one is about fixing the way to calculate the land fee.

mherzl:
@Auahi I was thinking the presence of the land fee is the underlying cause of this thread's concerns, but as you suggest here I'll write my ideas on the more specific topic of determining tax limits.

Yes, without forcing businesses to report their official economic activity it may be impossible to calculate gdp to a meaningful degree of accuracy, and as a result the tax limit may not be meaningful.

[...]

PR #419, opened 2016-05-08
removal of the land fee

mherzl:
my suggestion as per discussion:
https://github.com/liberland/constitution/issues/413#issuecomment-217731237


KacperZajc:
Land fee stays.
At least at this point;
This has been discussed on several occasions.

PR #429, opened 2016-05-17
Fixes

terrorist96:


KacperZajc:
I've changed this bit on the oath of secrecy and now I cannot merge your PR.

terrorist96:
Would you prefer your version or should I make another PR with my wording (all parties vs all Persons)?

KacperZajc:
I actually wanted to use 'all Parties' but then I realised that the term 'parties' traditionally refers only to prosecution/defence, not anyone adjudicating the dispute or working in the administration of the court.

terrorist96:
Really? I would think that 'all parties' would include anyone and everyone
who is involved (i.e. the Judge, the prosecutor and his team, the defense
and their team, the jury, a bailiff, a court stenographer, etc.)

KacperZajc:
Probably depends who you ask but this is not how I would construe this term so it is safer to use 'persons'.

Issue #418, opened 2016-05-08
V.4 video/audio recording


terrorist96:
> **§V.4.** No law shall prohibit video and/or audio recording of any Agent of the Public Administration whilst on duty in public space and all other areas where there is no reasonable expectation of privacy

I think she should clear this up. It can be interpreted in two different ways:
1. You can record an Agent while on duty if he is in a public space or any other area there is no privacy expectation; or
2. You can record an Agent while on duty if he is in a public space, and you can also record in general where there is no expectation of privacy.

Which interpretation are we going with? I'd prefer the second interpretation, so as to protect journalists' right to gather evidence for a story, or just a regular person recording some landscapes, etc.

If the fear is the government setting up cameras everywhere to watch everyone, we can prohibit this, but allow for private businesses to have security cameras. And if the government wants to investigate a crime, they can get a warrant to get a copy of the video footage from the owner.

Auahi:
I like the second definition as well, that being said I think it should be written in two seperate lines. And how would we stop government from filming everyone, is it possible to put something about that into the constitution?

PR #445, opened 2016-05-24
Right to record in public

terrorist96:
Addresses #418. Please detail any specific objections you may have.


KacperZajc:
could you add 'including Agents of Public Administration'
just to make it clear

terrorist96:
I could, but this would already be included since there's no exception for it. I thought you were looking to shorten the Constitution. :P

Issue #451, opened 2016-05-25
ID cards


KacperZajc:
§VIII.6. No Individual shall be compelled to identify himself or herself to any Agent of the Public Administration, unless upon arrest, a reasonable suspicion of a violation of the law or pursuant to a Warrant or Court order or subpoena; the Public Administration shall not issue any identification cards  for the Citizens of the Free Republic of Liberland other than passports.

Should we add that?

terrorist96:
I think the PA should be able to issue ID cards if it wants, like for voting, or purchasing alcohol/drugs that are prohibited to minors, or maybe even driving if we end up requiring a driver's license.

KacperZajc:
aren't those cards proving qualification rather than ID?

KacperZajc:
you might be right, this might too broad

Issue #441, opened 2016-05-23
Right to refuse to ID


terrorist96:
No Individual shall be compelled to identify himself or herself to any Agent of the Public Administration, unless upon arrest, an articulable suspicion of a violation of the law or pursuant to a Warrant.

KacperZajc:
No Individual shall be compelled to identify himself or herself to any Agent of the Public Administration, unless upon arrest, a reasonable suspicion of a violation of the law or pursuant to a Warrant or a Court order.

And we add this provision to the Privacy Rights Article of the Bill of Rights.
But I dont promise there are no more exceptions to this rights.

terrorist96:
> No Individual shall be compelled to identify himself or herself to any Agent of the Public Administration, unless upon arrest, a reasonable suspicion of a violation of the law or pursuant to a Warrant or ~~a~~ Court order.

liberlandcitizen:
Why should you have to identify yourself just because a police officer has
"reasonable suspicion?" Cops lie about this all of the time and use it to
harass people.

terrorist96:
Reasonable suspicion requires an articulate set of facts. If he lies about it, then he arrests you, you get out, sue his ass and win.

liberlandcitizen:
​It is often very difficult to prove that an officer lied. You can't
disprove what they what they claim they saw, heard or smelled. These types
of laws/rules do nothing to protect society. In a free nation, a free
person shouldn't have to say ANYTHING to the police if they don't want to.
At least not until they are arrested and only then their name and address.

I think an officer should be able to temporarily detain somebody if they
have reasonable suspicion, but your right to remain silent should not
evaporate just because they suspect something.​

terrorist96:
> but your right to remain silent should not
evaporate just because they suspect something.​

It doesn't. You only need to identify yourself if you're suspected of committing a crime, so they can write you a citation if need be or check you for any outstanding warrants. None of this happens unless there's suspicion of a crime.

liberlandcitizen:
If a citation is issued or an arrest has been made, that moves beyond
suspicion. ​This means an officer has evidence of a crime. At that point, I
think it's fair for citizens to have to identify themselves by name, and by
address. Nothing more, nothing less. However no citizen should have to
identify his or herself just because an officer has *suspicion*. Police can
use this to abuse citizens, or to fish for arrests, people with warrants,
etc. Example from my country: police demanding id from people of a certain
color. Their skin color makes them suspicious they are here illegally.
[...]

PR #455, opened 2016-05-27
reasonable suspicion -> probable cause

terrorist96:


KacperZajc:
see #441

KacperZajc:
please update it as per #451 before I merge

KacperZajc:
what's valid arrest?

terrorist96:
I mentioned in the summary. 

> Added valid, so as to protect people for refusing to identify when it's not a valid (unconstitutional/unlawful) arrest.

PR #490, opened 2016-06-20

terrorist96:
Changed wording so that a sentence can only be community work. "As part of" implies there's other parts to the sentence.


KacperZajc:
it can be community work and a fine, there might be other parts of the sentence too

KacperZajc:
Could you explain what is wrong with the current wording? 
I dont really get it.

terrorist96:
"as a part of a sentence" implies that it's only a part of a sentence, and not the whole sentence. My wording allows for a sentence to consist of community service only OR community service, and other things (fines, etc.). Your wording limits it to community service and other things. A sentence of ONLY community service would be impermissible under your wording. I'm simply allowing for the option by keeping the language more neutral.

KacperZajc:
ok if you say so

PR #527, opened 2016-09-03
fix

terrorist96:
Removed 'scheduled and completed' because it broke the sentence structure


terrorist96:
Also, this is the only provision which contains a serial comma whereas it's been avoided in all other provisions. I don't think it can be avoided due to the wording of this provision and it breaks the consistency. :disappointed:

KacperZajc:
I think this part is important, can you come up with wording which would encompass this bit yet sound better than mine?

KacperZajc:
how about:

§VIII.5. The Public Administration shall (a) record and transmit video and audio of, (b) publish and archive documents of, and (c) render available to the general public to attend, all affairs of the Assembly and all proceedings before the Courts other than those concerned with Warrants, **both scheduled and completed**; all documents officially produced by any Member of the Cabinet shall also be published within twenty-four hours; nothing in this provision shall prevent the Public Administration from protecting information classified as a state secret, including the debate and vote on such a classification.

terrorist96:
@KacperZajc ready for merge

PR #537, opened 2016-09-27

terrorist96:


KacperZajc:
what's wrong with dividing §VII.1. into two separate provisions?
it is better for transparency

